PR #63, opened 2021-01-10
Issue33 commong function interface

j-c-cook:


MassimoCimmino:
This was pushed to `issue33_CommongFunctionInterface` instead.

PR #76, opened 2021-03-02
Issue33 commong function interface

j-c-cook:


MassimoCimmino:
I have some work in progress on this branch. I will push some commits, hopefully later today, and see what kind of conflicts arise.

I took the time to implement a solver class that eliminates duplicate code that we find in each of the boundary conditions.

pygfunction/gfunction.py
```diff
@@ -12,13 +12,18 @@
 from .heat_transfer import thermal_response_factors, finite_line_source
 from .networks import Network, network_thermal_resistance
 
+
 class gFunction(object):
-    def __init__(self, boreholes_or_network, alpha, time=None,
-                 method='similarities', boundary_condition=None, options=None):
```
j-c-cook, line 17: @MassimoCimmino You have method here as a string. Is `use_similarities` more than a True/False? I would understand having a `str` for the input if there's more than just two options or there will be.

j-c-cook:
Do you intend to get rid of the `EIFT`, `UHF`, etc. stand alone methods?

pygfunction/gfunction.py
MassimoCimmino: I expect there will be more than two options in the near future. We have some upcoming work that might be of interest to pygfunction.

MassimoCimmino:
I say it would be a good idea to keep the existing functions for the time being for backwards compatibility, perhaps with a deprecation warning? (if future development is only to be made on the gFunction class).

pygfunction/gfunction.py
j-c-cook: Great, then I'll need to revert back to using type `str` for the `method`.

One thing though, there is a `method` in `options={}` because you can specify `linear`, etc. for the `scipy.interp` stuff.

j-c-cook:
That sounds good to me. Today, I am not a proponent for deprecation of the stand alone functions. I think it would be **nice** if the functions could have minimized repetition (like what you've done today}, but also the ability to call each one independently.

MassimoCimmino:
Another option is to call the new gFunction class in the dedicated functions, for example:

```
def uniform_temperature(boreholes, time, alpha, nSegments=12, method='linear',
                        use_similarities=True, disTol=0.01, tol=1.0e-6,
                        processes=None, disp=False):

    if use_similarities:
        method = 'similarities'
    else:
        method = 'detailed'
    boundary_condition = 'UBWT'
    options = {'nSegments':nSegments,
                      'disTol':disTol,
                      'tol':tol,
                      'processes':processes,
                      'disp':disp}
    gfunc = gFunction(boreholes, alpha, time=time, options=options)
    return gfunc.gFunc
```

pygfunction/gfunction.py
MassimoCimmino: The interpolation option could be replaced by `kind` to align with `scipy` terminology?

j-c-cook: Yes replacing `method` with `kind` for the `scipy.interp1d` option sounds good to me.

j-c-cook:
That would be vastly superior to deprecation. 

Before modifying the stand alone functions we need to make sure they are returning the same g-function curves as what is in the new gFunction class.

MassimoCimmino:
Here is a comparison of the new and old interfaces for all rectangular fields from 1 by 1 to 12 by 12 calculated from t = 100 h to t = 3000 years. The differences seem negligible.

Figure_4
Figure_5

```
from __future__ import absolute_import, division, print_function

import matplotlib.lines as mlines
import matplotlib.pyplot as plt
import numpy as np
from matplotlib.ticker import AutoMinorLocator

import pygfunction as gt


def main():
    # Simulation parameters

    # Borehole dimensions
    D = 4.0             # Borehole buried depth (m)
    H = 150.0           # Borehole length (m)
    r_b = 0.075         # Borehole radius (m)
    B = 7.5             # Borehole spacing (m)

    # Thermal properties
    alpha = 1.0e-6      # Ground thermal diffusivity (m2/s)

    # g-Function calculation options
    nSegments = 12
    disp = True
    options = {'nSegments':nSegments, 'disp':disp}

    # Geometrically expanding time vector.
    dt = 100*3600.                  # Time step
    tmax = 3000. * 8760. * 3600.    # Maximum time
    Nt = 25                         # Number of time steps
    ts = H**2/(9.*alpha)            # Bore field characteristic time
    time = gt.utilities.time_geometric(dt, tmax, Nt)
    
    # g-Function calculation
    N_1_max = 12
    N_2_max = 12
    N_gfunc = int(N_1_max*(N_2_max + 1)/2)
    gfunc_new = np.zeros((N_gfunc, Nt))
    gfunc_old = np.zeros((N_gfunc, Nt))

    i = 0
    for N_1 in range(1, N_1_max + 1):
        for N_2 in range(N_1, N_2_max + 1):
            field = gt.boreholes.rectangle_field(N_1, N_2, B, B, H, D, r_b)
            gfunc_new[i,:] = gt.gfunction.gFunction(field, alpha, time=time, options=options).gFunc
            gfunc_old[i,:] = gt.gfunction.uniform_temperature(field, time, alpha, nSegments=nSegments, disp=disp)
            i += 1

    # Initialize figure

    plt.rc('figure')
    fig = plt.figure()
    ax1 = fig.add_subplot(111)
    ax1.plot(gfunc_old, gfunc_new, 'k.')
    # Axis labels
    ax1.set_ylabel('$g (new)$')
    ax1.set_xlabel('$g (old)$')
    # Axis limits
    # ax1.set_xlim([-10.0, 5.0])
    # ax1.set_ylim([0., 70.])
    # Show minor ticks
    ax1.xaxis.set_minor_locator(AutoMinorLocator())
    ax1.yaxis.set_minor_locator(AutoMinorLocator())
    # Adjust to plot window
    plt.tight_layout()

    plt.rc('figure')
    fig = plt.figure()
    ax1 = fig.add_subplot(111)
    ax1.plot(gfunc_old, gfunc_new-gfunc_old, 'k.')
    # Axis labels
    ax1.set_ylabel('$Delta-g$')
    ax1.set_xlabel('$g (old)$')
    # Show minor ticks
    ax1.xaxis.set_minor_locator(AutoMinorLocator())
    ax1.yaxis.set_minor_locator(AutoMinorLocator())
    # Adjust to plot window
    plt.tight_layout()

    return


# Main function
if __name__ == '__main__':
    main()
```

MassimoCimmino:
Closed to open a PR from the origin branch

PR #77, opened 2021-03-02
Properties module and fluid object

j-c-cook:
Here's my draft of the `Fluid` object. I think it might be cleaner to pass the `Fluid` object to methods or objects that make use of the fluid properties.


doc/source/modules.rst
```diff
@@ -14,3 +14,4 @@ Modules
    networks
    pipes
    utilities
+   properties
```
MassimoCimmino, line 17: An alternative to `properties` could be `fluid` (it contains properties for fluids) or `media` (if it ever grows to include soil/grout as well).

pygfunction/properties.py
```diff
@@ -0,0 +1,122 @@
...
+class Fluid:
+    """
+        An object for handling the fluid properties
+
+        Parameters
+        ----------
+        m_flow: float
+            The fluid mass flow rate per borehole (kg/s)
+        T: float, optional
+            The temperature of the fluid (Kelvin)
```
MassimoCimmino, line 16: With the implementation of the `Fluid` class, we should settle on a unit system convention.

My preference is : degC, J, kg, Pa, s. This decision should be added in the _Styleguide_ section of [CONTRIBUTING.md](https://github.com/MassimoCimmino/pygfunction/blob/master/CONTRIBUTING.md).

pygfunction/properties.py
```diff
@@ -0,0 +1,122 @@
...
+    def dynamic_viscosity(self):
+        return PropsSI('V', 'T', self.T, 'P', self.P, self.fluid_mix)
+
+    def kinematic_viscosity(self):
+        return self.mu / self.rho
+
+    def mass_heat_capacity(self):
+        return PropsSI('C', 'T', self.T, 'P', self.P, self.fluid_mix)
+
+    def volumetric_heat_capacity(self):
```
MassimoCimmino, line 115: See the other comment. Methods in the 'Pipe' and 'Network' classes use J/kg-degC.

pygfunction/properties.py
```diff
@@ -0,0 +1,122 @@
...
+        >>> mix = 'MMA'
+        >>> percent = 5
+        >>> fluid = gt.properties.Fluid(m_flow, T=T, P=P, mixer=mix, percent=percent)
+        >>> print(fluid.__repr__())
+
+        >>> # ethanol / water
+        >>> mix = 'MEA'
+        >>> percent = 10
+        >>> fluid = gt.properties.Fluid(m_flow, T=T, P=P, mixer=mix, percent=percent)
+        >>> print(fluid.__repr__())
```
MassimoCimmino, line 65: Is there a difference between `print(fluid.__repr__())` and `print(fluid)`?

j-c-cook: If the `__repr__()` function did not exist, then `print(fluid)` has the following output:
```
<pygfunction.properties.Fluid object at 0x000001D5B26FFAC8>
```
However, with the `__repr__(self)` method uncommented, `print(fluid)` has this output:
```
<class 'pygfunction.properties.Fluid'>
P = 239220.0
Pr = 7.070547362075505
T = 293.0
cp = 4186.2937048371
fluid_mix = INCOMP::MEG-0%
k = 0.5988756481542833
mu = 0.0010114862770798584
nu = 1.0141456799573551e-06
rho = 997.3776914598615
rhoCp = 4175.315951103378
```
The next commit will just have `print(fluid)` because with the `__repr__` method defined, they result in the same.

j-c-cook:
The most recent commit (a91a78f9fea9a079e81de5a3a61b5dd346f3d91a) does the following:

- Removes m_dot from the fluid properties object
- Takes in temperature in Celsius 
- changes `print(fluid.__repr()__)` to `print(fluid)` in the examples

doc/source/modules.rst
j-c-cook: `media` is fine with me. I'd prefer that to `fluid` for if we do get around to including other objects.

pygfunction/properties.py
j-c-cook: Mass flow rate has been removed because this object is going to be used as a method for quick access to fluid properties, not an object to be passed around inside of `pygfunction`.

The temperature passed into the function is now in Celsius.

MassimoCimmino:
I submitted a PR on your fork [#1](https://github.com/j-c-cook/pygfunction/pull/1) for this issue.

j-c-cook:
I added the ability to access pure `Water` in the `Fluid` object and resolved conflicts from `master`. The conflicts were encountered because issue #81 was merged in pull request #82.

It appears to be ready to merge to me.

MassimoCimmino (approved):

PR #79, opened 2021-03-20
Issue36 coaxial pipes

j-c-cook:
In progress

### DIFF
--- a/pygfunction/pipes.py
+++ b/pygfunction/pipes.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import numpy as np
 from scipy.constants import pi
 from scipy.special import binom
+from scipy.interpolate import interp1d
 
 
 class _BasePipe(object):
@@ -2010,6 +2011,55 @@ def fluid_friction_factor_circular_pipe(m_flow, r_in, visc, den, epsilon,
     return fDarcy
 
 
+def Gnielinski(Re, Pr, fDarcy):
+    """
+    Cengel and Ghajar (2015, pg. 497) say that the Gnielinski (1976) equation should be considered the preferred
+    equation for determining the Nusselt number for the transition and turbulent in internal forced flow
+
+    .. math::
+
+        0.5 \leq Pr \leq 2000
+
+        3 \times 10^3 < Re < 5 \times 10^6
+
+
+    Parameters
+    ----------
+    Re : float
+        Reynolds number
+    Pr : float
+        Prandlt Number
+    fDarcy : float
+        Darcy-Weisbach friction factor
+
+    Returns
+    -------
+    Nu : float
+        The Nusselt number
+
+    """
+    import warnings
+
+    # Warn the user if the Reynolds number is out of bounds, but don't break
+    if 3.0E03 < Re < 5.0E06:
+        pass
+    else:
+        warnings.warn('This Nusselt calculation is only valid for Reynolds number in the range'
+                      'of 3.0E03 < Re < 5.0E06, your value falls outside of the range at'
+                      ' Re={0:.4f}'.format(Re))
+    # Warn the user if the Prandlt number is out of bounds
+    if 0.5 <= Pr <= 2000.:
+        pass
+    else:
+        warnings.warn('This Nusselt calculation is only valid for Prandlt numbers in the range'
+                      'of 0.5 <= Pr <= 2000, your value falls outside of the range at'
+                      ' Pr={0:.4f}'.format(Pr))
+
+    Nu = 0.125*fDarcy * (Re - 1.0e3) * Pr / \
+         (1.0 + 12.7 * np.sqrt(0.125*fDarcy) * (Pr**(2.0/3.0) - 1.0))
+    return Nu
+
+
 def convective_heat_transfer_coefficient_circular_pipe(m_flow, r_in, visc, den,
                                                        k, cp, epsilon):
     """
@@ -2054,15 +2104,33 @@ def convective_heat_transfer_coefficient_circular_pipe(m_flow, r_in, visc, den,
     # Darcy friction factor
     fDarcy = fluid_friction_factor_circular_pipe(m_flow, r_in, visc, den,
                                                  epsilon)
-    if Re > 2300.:
+
+    # To ensure there are no dramatic jumps in the equation, an interpolation in a transition
+    # region of 2300 <= Re <= 4000 will be used
+    # Cengel and Ghajar (2015, pg. 476) state that Re> 4000 is a conservative value to consider
+    # the flow to be turbulent in piping networks
+
+    transition_lower = 2300.
+    transition_upper = 4000.
+
+    if Re > transition_upper:
         # Nusselt number from Gnielinski
-        Nu = 0.125*fDarcy * (Re - 1.0e3) * Pr / \
-            (1.0 + 12.7 * np.sqrt(0.125*fDarcy) * (Pr**(2.0/3.0) - 1.0))
+        Nu = Gnielinski(Re, Pr, fDarcy)
+    elif transition_lower < Re < transition_upper:
+        Nu_lam = 3.66  # constant surface temperature laminar Nusselt number (Re = 2300.)
+        # Nusselt number at the upper bound of the "transition" region between laminar value
+        # and Gnielinski correlation (Re = 4000.)
+        Nu_turb = Gnielinski(transition_upper, Pr, fDarcy)
+        # outer bound point pairs for 1D linear interpolation
+        interp_pts = [(transition_lower, Nu_lam), (transition_upper, Nu_turb)]
+        x, y = list(zip(*interp_pts))  # seperate x and y values
+        f = interp1d(x, y, kind='linear')
+        Nu = f(Re).tolist()  # scipy.interp1d returns a value in an array, call tolist() to make it a float value
     else:
         Nu = 3.66
     h_fluid = k * Nu / D
 
-    return h_fluid
+    return h_fluid, Re, Nu
 
 
 def conduction_thermal_resistance_circular_pipe(r_in, r_out, k):


j-c-cook:
With my most recent commit (9022abc68d614a1c601237ce6126614afd4d8e50) I have modified the function that calculates the  [convective heat transfer coefficient in a circular pipe][1] with forced flow. After some discussion with J.D. Spitler, he said that for design purposes it can be important to have smooth transitions from laminar to turbulent, which can be provided via interpolation. Cengel and Ghajar (2015, pg. 476) state that considering full turbulence in piping networks for any Re > 4000 is conservative. I have incorporated a "transition" 1D interpolation between the laminar (Nu=3.66) and the Nu(Re=4000)) while 2300 < Re < 4000. 

By varying the mass flow rate and holding all other values constant, I plotted the Reynolds vs. the Nusselt number. [jump.pdf](https://github.com/MassimoCimmino/pygfunction/files/6174837/jump.pdf) plots the results of what the function was, and shows that there was a sharp jump from 3.66 to about 20 for the Nusselt number when the value of the Reynolds number became greater than 2300. The new function with interpolation from 2300 < Re < 4000 creates [no-jump.pdf](https://github.com/MassimoCimmino/pygfunction/files/6174840/no-jump.pdf).


[1]: https://github.com/j-c-cook/pygfunction/blob/584754b86234ad4b516618813971548741eec339/pygfunction/pipes.py#L1941

MassimoCimmino:
The range [2300, 4000] is also found elsewhere in the literature, e.g. [Gnielinski (2013)](https://www.sciencedirect.com/science/article/pii/S0017931013003207).
--------------------
Gnielinski, V. (2013). On heat transfer in tubes. _International Journal of Heat and Mass Transfer_, 63, 134-140.

j-c-cook:
Commit 843dd7e881db6c5e6856f7405bf6aff46a1e93dc provides no functionality difference. The linear interpolation is no longer done by `scipy`, but uses Equations 16 and 17 from Gnielinski (2013). Documentation is enhanced for Gnielinksi() and convective_heat_transfer_coefficient_circular_pipe(); references are included.

j-c-cook:
Hellstrom's inner and outer Nusselt numbers are used for laminar flow. The transition from 2300 < Re < 4000 makes use of the linear interpolation from Gnielinski (2013) for both inner and outer. Cengel and Ghajar (2015) state that inner and outer Nusselt numbers are equivalent for turbulent flow, and that  Gnielinski (1975) can be used. 

A plot of Reynolds vs. Nusselt is attached. 


[Re_vs_Nu_annulus.pdf](https://github.com/MassimoCimmino/pygfunction/files/6261206/Re_vs_Nu_annulus.pdf)

j-c-cook:
@MassimoCimmino Could you take a look at this and provide feedback if you'd like something changed?

I still need to do some further verification before its finalized, but the methodology is there. Now would be a good time to know if you'd like to see something different. Whenever you get the chance, thanks.

MassimoCimmino:
I like the idea of the new borehole thermal resistance function. Here are my first thoughts:

- The matrix of thermal resistances `SingleCoaxialPipe._Rd` needs to be evaluated in `__init__` to enable all functionalities of the pipe classes, which are needed for the 'MIFT' boundary condition in `gFunction` and for the `Network` class. The effective `Rb` (uniform temperature) could then be evaluated from the existing `borehole_thermal_resistance()` (This function is a generalization of the Hellström method to *N* pipes).
- On that same topic, there should be a way to obtain an expression for the uniform heat flow resistance from the `R_delta` matrix so that the new thermal resistance calculations are compatible with any `pipe` object. I will look into this.
- There are implications in implementing the 'UHTR' and 'AVG' effective thermal resistances. The *effective* borehole wall temperature calculated using the 'MIFT' relies on the 'UBWT' resistance to find an *effective* borehole wall temperature (as defined in [1]). This is not a huge problem as long as the `Network` class is used to evaluate fluid temperatures during simulations (even for simple fields with parallel-connected boreholes). We need to make sure the default parameters do not break the implementation of the `gFunction` or `Network` classes.
- While it is not ideal, the existing `pipe` classes do not calculate `R_f` internally but instead take it as an input. We could consider opening a new issue for this change. This also provides opportunities to implement temperature dependent physical properties in the `pipe` and `Network` classes using the new `media` module.
- Using the same `r_in` and `r_out` arrays as in other classes may be preferable to avoid implementing a new `visualize_pipes()` method (I know the variable definition are from my commit).
- A new `_check_geometry()` method is required.

------

[1] Cimmino, M. (2019). Semi-Analytical Method for g-Function Calculation of bore fields with series-and parallel-connected boreholes. Science and Technology for the Built Environment, 25(8), 1007-1022.

j-c-cook:
The `thermal_resistances_coaxial` function takes in one pipe resistance `R_p` value, how will we account for the inner and outer pipe having a different resistance?
Given that we now have pipes with different radius, how will we handle that the existing `visualize_pipes()` only considers an `r_in` and `r_out`?
Do you want a new `_check_geometry()` method or to modify the existing?

MassimoCimmino:
- I did not think of the different `R_p` values. I cannot think of an elegant way to implement it in a convenient interface. It might be a good idea to move ahead with opening a new issue and handle the calculation of film resistances at the initialization of `pipe` objects instead of having `R_fp` as an input. It will make the interface to the different pipes consistent, and v2.0.0 is the best time to change the interface in a non-backwards-compatible way.
- I got confused here and misremembered that `r_in` and `r_out` were arrays (they are not). The current `visualize_pipes()` method can be adapted to cover both cases (with `r_in` and `r_out` as arrays in coaxial pipes). The code below creates `r_in` and `r_out` arrays when floats or scalars are provided:

```
    def visualize_pipes(self):
        """
        Plot the cross-section view of the borehole.

        Returns
        -------
        fig : figure
            Figure object (matplotlib).

        """
        import matplotlib.pyplot as plt
        from matplotlib.ticker import AutoMinorLocator

        # Initialize figure
        LW = .5    # Line width
        FS = 12.    # Font size

        plt.rc('figure', figsize=(80.0/25.4, 80.0/25.4))
        fig = plt.figure()
        ax = fig.add_subplot(111)

        # Borehole wall outline
        borewall = plt.Circle((0., 0.), radius=self.b.r_b,
                              fill=False, linestyle='--', linewidth=LW)
        ax.add_patch(borewall)

        # Pipes
        r_in = np.atleast_1d(self.r_in)
        if len(r_in) == 1: r_in = np.repeat(r_in, 2*nPipes)
        r_out = np.atleast_1d(self.r_out)
        if len(r_out ) == 1: r_out = np.repeat(r_out , 2*nPipes)
        for i in range(self.nPipes):
            # Coordinates of pipes
            (x_in, y_in) = self.pos[i]
            (x_out, y_out) = self.pos[i + self.nPipes]

            # Pipe outline (inlet)
            pipe_in_in = plt.Circle((x_in, y_in), radius=r_in[i],
                                    fill=False, linestyle='-', linewidth=LW)
            pipe_in_out = plt.Circle((x_in, y_in), radius=r_out[i],
                                     fill=False, linestyle='-', linewidth=LW)
            ax.text(x_in, y_in, i + 1,
                    ha="center", va="center", size=FS)

            # Pipe outline (outlet)
            pipe_out_in = plt.Circle((x_out, y_out), radius=r_in[i + self.nPipes],
                                     fill=False, linestyle='-', linewidth=LW)
            pipe_out_out = plt.Circle((x_out, y_out), radius=r_out[i + self.nPipes],
                                      fill=False, linestyle='-', linewidth=LW)
            ax.text(x_out, y_out, i + self.nPipes + 1,
                    ha="center", va="center", size=FS)

            ax.add_patch(pipe_in_in)
            ax.add_patch(pipe_in_out)
            ax.add_patch(pipe_out_in)
            ax.add_patch(pipe_out_out)

        # Configure figure axes
        ax.set_xlabel('x (m)')
        ax.set_ylabel('y (m)')
        plt.axis('equal')
        ax.xaxis.set_minor_locator(AutoMinorLocator())
        ax.yaxis.set_minor_locator(AutoMinorLocator())
        plt.tight_layout()

        return fig
```

- A new `_check_geometry()` is required, since the current one checks for collision between pipes. It also checks the other inputs.

j-c-cook:
@MassimoCimmino I have modified [custom_borehole.py](https://github.com/j-c-cook/pygfunction/blob/issue36_CoaxialPipes/pygfunction/examples/custom_borehole.py) to include Double U-Tubes in series and parallel. Once we have worked out the Coaxial interface, we could add Coaxial to the example as well. The custom borehole example would then serve the purpose of seeing how to define any tube arrangement available in pygfunction all in one condensed location.

MassimoCimmino:
I opened two issues, #120 and #121, to divide the different topics of this PR as it seems several points are out of scope:

- The modification of the `borehole_thermal_resistance` is moved to #121;
- The internal calculation of fluid and pipe thermal resistance is more of a long-term effort and requires to refactor the interfaces to all `pipe` classes. It is moved to #120.

This way we can quickly have the base functionality of the `coaxial` class and update the interface (and capabilities) of all `pipe` class at the same time so there is no disparity between classes.

All would be required to implement the `coaxial` class (assuming it can inherit from the `SingleUTube` class) is an `__init__` method with the necessary inputs to build `self._Rd`, and if required `_check_geometry` and `visualize_pipes` methods.

j-c-cook:
> All would be required to implement the `coaxial` class (assuming it can inherit from the `SingleUTube` class) is an `__init__` method with the necessary inputs to build `self._Rd`, and if required `_check_geometry` and `visualize_pipes` methods.

Why does `class Coaxial` need to inherit from `SingleUTube`, why not just `_BasePipe`? Does Claesson and Hellstrom (2011) [thermal_resistances](https://github.com/MassimoCimmino/pygfunction/blob/7d37b8b3bf4b349701681e42281c21cccb54ca42/pygfunction/pipes.py#L1780-L1786) can handle coaxial pipes?

MassimoCimmino:
`SingleUtube` implements the solution of Hellström's thesis (1991) for the fluid temperature profiles. From page 95:
diff_eq

The heat transfer problem for a coaxial borehole should have the same form, with the thermal resistance of the inner pipe to the borehole wall being `R_d_1 -> inf` (or `R_d_2 -> inf` -- in *pygfunction* `1` is the inlet downward flowing pipe). `R_d_12` is replaced by the convective + conductive resistance between fluid volumes.

The remaining resistance (either `R_d_2` or `R_d_1`) can be evaluated using the multipole method. The line source approximation is correct for a centrally located pipe but the multipole method would be required if the pipes are off-centered.

This could be implemented by simply inheriting from `SingleUtube` and re-defining `__init__` to replace the current call to `thermal_resistance` by the correct definition of the thermal resistance matrix.

https://github.com/MassimoCimmino/pygfunction/blob/7d37b8b3bf4b349701681e42281c21cccb54ca42/pygfunction/pipes.py#L824-L843


--------------


To handle some flexibility in the definition of the inlet/outlet pipe. A way to do it would be to keep a similar call to the function as for `SingleUTube` but have `r_in` and `r_out` as arrays, e.g. :
```
def __init__(self, pos, r_in, r_out, borehole, k_s, k_g, R_ff, R_fp, J=2):
```
with
```
pos = (x, y) # Coordinates of the coaxial pipe axis
r_in = np.array([r_in_inlet, r_in_outlet]) # Inner radius of the pipes, starting from the inlet pipe
r_out = np.array([r_out_inlet, r_out, outlet]) # Outer radius of the pipes, starting from the inlet pipe
R_ff : float # Fluid-to-fluid thermal resistance between the two fluid volumes
R_fp : float # Fluid-to-outer-pipe-wall thermal resistance of the outer pipe in contact with the grout
```

Then, the values of `_Rd[0,0]` and `_Rd[1,1]` can be determined depending on if `r_in_inlet > r_in_outlet` is `True` of `False`, and `_Rd[0,1]=_Rd[1,0]=R_ff`.

j-c-cook:
It'll help to set this up the way you've described. I'm going to take a crack at it tomorrow. This would be a nice addition.

j-c-cook:
Pipe positioning providing some issues. There's a divisible by 0 error in dz. 

https://github.com/j-c-cook/pygfunction/blob/70b65018b493199d4f025a45e4621cb5683a00b5/pygfunction/pipes.py#L2282-L2303

j-c-cook:
Similar issue with the line source approximation. Could you take a look @MassimoCimmino ?

MassimoCimmino:
> Pipe positioning providing some issues. There's a divisible by 0 error in dz.
> 
> https://github.com/j-c-cook/pygfunction/blob/70b65018b493199d4f025a45e4621cb5683a00b5/pygfunction/pipes.py#L2282-L2303
> 
> @MassimoCimmino

@j-c-cook You only need to call `multpole()` for the outer pipe. This happens because it attempts to calculate the pipe-to-pipe resistance for 2 pipes at the same coordinates.

j-c-cook:
Setting `pos=(0., 0.)` breaks with:
```
  File "/media/jackcook/SHARED/Masters/Research/pygfunction/pygfunction/pipes.py", line 1010, in _update_model_variables
    self._beta2 = 1./(self._Rd[1][1]*m_flow_in[0]*cp_in[0])
IndexError: index 1 is out of bounds for axis 0 with size 1
```
And setting `pos=(0., 0.)` breaks with:
```
  File "/media/jackcook/SHARED/Masters/Research/pygfunction/pygfunction/pipes.py", line 2282, in <listcomp>
    z_p = np.array([pos[i][0] + 1.j*pos[i][1] for i in range(n_p)])
TypeError: 'float' object is not subscriptable
```

What do you think?

MassimoCimmino:
Do you have a script I can run?

j-c-cook:
I'm working out of the `customborehole.py` example:

https://github.com/j-c-cook/pygfunction/blob/8224783ba11e1ec4520da9624cd540ce7d6ded45/pygfunction/examples/custom_borehole.py#L64-L107

The example file has changes in this branch.

MassimoCimmino:
I pushed some changes to my branch:
https://github.com/MassimoCimmino/pygfunction/blob/844a187358efc83ec3d4795d8bdb899d5e188af9/pygfunction/pipes.py#L1782-L1811

The idea is to overwrite the `__init__` of `SingleUTube` and build the `_Rd` matrix from the new inputs. `thermal_resistances` is only called fore the outer pipe in contact with the grout.

j-c-cook:
Thank you. Merged. I'll work to clean up what's there and validate borehole resistance.

j-c-cook:
@MassimoCimmino See this plot of Single and double U-tube and coaxial effective borehole resistances plotted versus Reynolds number. The Reynolds number being used for the Coaxial is the Reynolds number in the annulus. 

What do you think about these results?

Validation

j-c-cook:
Physically, does the plot above (https://github.com/MassimoCimmino/pygfunction/pull/79#issuecomment-904717004) make sense?

MassimoCimmino:
Are these results for the same geometry as in `custom_borehole.py` for the three tools? Also, for `GLHEDT` and `GLHEPRO`, is this the *effective* borehole resistance (usually called `R_b^star` and different from `R_b`) with a uniform borehole wall temperature?

When running `custom_borehole.py`, the `GLHEDT` and `GLHEPRO` curves look like the cross-section borehole resistance given by `print(Coaxial._Rd[1,1])`.

You can get a sense of the *effective* thermal resistance by drawing the temperature profiles, e.g. :

```
    # Plot fluid temperature profiles
    nz = 20
    z = np.linspace(0., H, num=nz)
    T_b = 5.
    T_f_in = 10.0
    T_f_single = SingleUTube.get_temperature(
        z, T_f_in, T_b,  m_flow_borehole, fluid.cp)
    T_f_out_single = SingleUTube.get_outlet_temperature(
        T_f_in, T_b, m_flow_borehole, fluid.cp)
    T_f_coaxial = Coaxial.get_temperature(
        z, T_f_in, T_b,  m_flow_borehole, fluid.cp)
    T_f_out_coaxial = Coaxial.get_outlet_temperature(
        T_f_in, T_b, m_flow_borehole, fluid.cp)

    # Configure figure and axes
    fig = gt.utilities._initialize_figure()

    ax3 = fig.add_subplot(111)
    # Axis labels
    ax3.set_xlabel(r'Temperature [degC]')
    ax3.set_ylabel(r'Depth from borehole head [m]')
    gt.utilities._format_axes(ax3)

    # Plot temperatures
    ax3.plot(np.array([T_b, T_b]), np.array([0., H]), 'k--')
    ax3.plot(T_f_single, z, 'b-')
    ax3.plot(0.5*T_f_in + 0.5*np.array([T_f_out_single, T_f_out_single]), np.array([0., H]), 'b:')
    ax3.plot(T_f_coaxial, z, 'r-')
    ax3.plot(0.5*T_f_in + 0.5*np.array([T_f_out_coaxial, T_f_out_coaxial]), np.array([0., H]), 'r:')
    # ax3.legend(['Borehole wall', 'Single', 'Coaxial'])

    # Reverse y-axes
    ax3.set_ylim(ax3.get_ylim()[::-1])
    # Adjust to plot window
    plt.tight_layout()
```

Figure_7
`k--` is the Borehole wall temperature
`b-` is the `SingleUTube` temperature profile
`b:` is the `SingleUTube` arithmetic mean temperature (`0.5 * (T_f_in + T_f_out)`)
`r-` is the `Coaxial` temperature profile
`r:` is the `Coaxial` arithmetic mean temperature (`0.5 * (T_f_in + T_f_out)`)

j-c-cook:
The same geometry is used in Coaxial, but not the single U-tube. I can get that updated though. 

Yes, the GLHEDT coaxial is effective borehole thermal resistance (`R_b_star`) using the following code:

```
def compute_effective_borehole_resistance(self, m_flow, visc_f, den_f,
                                              k_f, cp_f, case, disp=False):
        # Inner pipe thermal resistance
        R_p_in = conduction_thermal_resistance_circular_pipe(self.r_in_in,
                                                             self.r_in_out,
                                                             self.k_p_in)
        if disp:
            print('Inner pipe thermal resistance, R_p_in: {}'.format(R_p_in))
        # Outer pipe thermal resistance
        R_p_out = conduction_thermal_resistance_circular_pipe(self.r_out_in,
                                                              self.r_out_out,
                                                              self.k_p_out)
        if disp:
            print('Outer pipe thermal resistance, R_p_out: {}'.format(R_p_out))
        # Grout thermal resistance
        R_grout = conduction_thermal_resistance_circular_pipe(self.r_out_out,
                                                              self.b.r_b,
                                                              self.k_g)
        if disp:
            print('Grout thermal resistance: {}'.format(R_grout))
        h_fluid_a_in, h_fluid_a_out, Re =\
            convective_heat_transfer_coefficient_concentric_annulus(m_flow,
                                                                self.r_in_out,
                                                                self.r_out_in,
                                                                visc_f,
                                                                den_f,
                                                                k_f,
                                                                cp_f,
                                                                self.eps_in)
        if disp:
            print('Annulus region, h_fluid_a_in = {}\th_fluid_a_out={}\tRe={}'
                  .format(h_fluid_a_in, h_fluid_a_out, Re))
        # Inner fluid convective resistance
        R_f_a_in = 1. / (h_fluid_a_in * 2 * pi * self.r_in_out)
        if disp:
            print('Inner fluid convective resistance, R_f_a_in: {}'.
                  format(R_f_a_in))
        # Outer fluid convective resistance
        R_f_a_out = 1. / (h_fluid_a_out * 2 * pi * self.r_out_in)
        if disp:
            print('Outer fluid convective resistance, R_f_a_out: {}'.
                  format(R_f_a_out))

        h_fluid_in = convective_heat_transfer_coefficient_circular_pipe(m_flow,
                                                           self.r_in_in,
                                                           visc_f,
                                                           den_f,
                                                           k_f,
                                                           cp_f,
                                                           self.eps_in)
        if disp:
            print('Inside convection coefficient, h_fluid_in: {}'.
                  format(h_fluid_in))
        R_f_in = 1. / (h_fluid_in * 2 * pi * self.r_in_in)
        if disp:
            print('Fluid thermal resistance, R_f_in: {}'.format(R_f_in))
        Ra = R_f_in + R_p_in + R_f_a_in
        if disp:
            print('Ra = {}'.format(Ra))
        R_12 = Ra
        Rb = R_f_a_out + R_p_out + R_grout
        if disp:
            print('Rb = {}'.format(Rb))

        Rb_star = borehole_thermal_resistance_2(Rb, Ra, R_12, self.b.H,
                                                self.r_in_in, m_flow, den_f,
                                                cp_f, case=case, disp=disp)
        return Rb_star
```

```
def borehole_thermal_resistance_2(Rb, Ra, R_12, H, r_in, m_flow, den, cp,
                                  case='AVG', disp=False):
    """
    Compute the effective borehole thermal resistance using the equations
    presented in Chapter 3 [#JavedSpitler2016]_ (Javed and Spitler 2016) of
    Advances in Ground-Source Heat Pump Systems [#Rees2016]_ (Edited by Rees
    2006).

    The uniform heat flux case (UHF) comes from Claesson and Hellström (2011)
    [#Claesson2011b]_:

    .. math::

        R_b^* = R_b + \\dfrac{1}{3R_a}\\bigg( \dfrac{H}{\\rho_f c_f V_f}
        \\bigg)^2

    The uniform borehole wall temperature case (UBHWT) also comes from
    Claesson and Hellström (2011) [#Claesson2011b]_:

    .. math::

        R_b^* = R_b \\eta coth \; \\eta

        \\eta = \dfrac{H}{\\rho_f c_f V_f} \\; \\dfrac{1}{2R_b}
        \\sqrt{1 + \dfrac{4R_b}{R_{1-2}}}

    Javed and Spitler (2016) [#JavedSpitler2016]_ state that neither the UHF
    or the UBHWT case are perfect representations of the physics, and that the
    mean of the two cases are often used in practice.

    Parameters
    ----------
    Rb : float
        Local borehole resistance (in m.K/W).
    Ra : float
        Total internal thermal resistance (in m.K/W).
    R_12 : float
        Direct coupling resistance (in m.K/W).
    H : float
        Height of borehole (in meters).
    r_in : float
        Inside pipe diameter (in meters)
    m_flow : float
        Fluid mass flow rate (in kg/s).
    den : float
        Fluid density (in kg/m3).
    cp : float
        Fluid specific isobaric heat capacity (in J/kg.K).
    case : str, optional
        - 'UHF': Uniform heat flux
        - 'UBHWT': Uniform borehole wall temperature
        - 'AVG': Average of UHF and UBHWT
        Default is 'AVG'.

    Returns
    -------
    Rb_star : float
        Effective borehole thermal resistance (in m.K/W)

    References
    ------------
    .. [#Rees2016] Rees, S. (2016). Advances in Ground-Source Heat Pump
        Systems: Vol. number 100. Elsevier Science & Technology.
    .. [#JavedSpitler2016] Javed, S., Spitler, J.D. (2016). Chapter 3 of
        Advances in Ground-Source Heat Pump Systems: Calculation of
        borehole thermal resistance. Pages 63-95, ISBN 9780081003114,
        https://doi.org/10.1016/B978-0-08-100311-4.00003-0
    """
    # Fluid velocity (for the inside pipe)
    V_flow = m_flow / den
    A_cs = pi * r_in ** 2
    V = V_flow / A_cs

    # The UHF case, effective borehole resistance, eq. (3.67)
    Rb_star_UHF = Rb + 1 / (3 * Ra) * (H / (den * cp * V)) ** 2
    if disp:
        print('The UHF effective borehole resistance (in m.K/W): {}'.
              format(Rb_star_UHF))

    # The UBH temperature case
    eta = H / (den * cp * V) * 1 / (2 * Rb) * \
          np.sqrt(1 + 4 * Rb / R_12)  # eq. (3.69)

    Rb_star_UBH = Rb * eta * (1 / np.tanh(eta))  # eq. (3.68)
    if disp:
        print('The UBH effective borehole resistance (in m.K/W): {}'.
              format(Rb_star_UHF))

    if case == 'UHF':
        return Rb_star_UHF
    elif case == 'UBHWT':
        return Rb_star_UBH
    elif case == 'AVG':
        return (Rb_star_UHF + Rb_star_UBH) / 2
    else:
        raise ValueError('The case provdided, {}, is not an option. The'
                         'options are UHF, UBHWT or AVG'.format(case))
```

j-c-cook:
I'm not saying what we have done here for Coaxial is _wrong_. I'd just like to know if based off the physics we think its _right_.

MassimoCimmino:
In these two instances in `borehole_thermal_resistance_2` :
```
    # The UHF case, effective borehole resistance, eq. (3.67)
    Rb_star_UHF = Rb + 1 / (3 * Ra) * (H / (den * cp * V)) ** 2
```
and
```
    eta = H / (den * cp * V) * 1 / (2 * Rb) * \
          np.sqrt(1 + 4 * Rb / R_12)  # eq. (3.69)
```

I believe `V` should be `V_flow`.

MassimoCimmino:
> I'm not saying what we have done here for Coaxial is _wrong_. I'd just like to know if based off the physics we think its _right_.

It is still possible that there is an issue with `Coaxial`. Any difference with the other tools should be looked at.

j-c-cook:
> In these two instances in `borehole_thermal_resistance_2` :
> 
> ```
>     # The UHF case, effective borehole resistance, eq. (3.67)
>     Rb_star_UHF = Rb + 1 / (3 * Ra) * (H / (den * cp * V)) ** 2
> ```
> 
> and
> 
> ```
>     eta = H / (den * cp * V) * 1 / (2 * Rb) * \
>           np.sqrt(1 + 4 * Rb / R_12)  # eq. (3.69)
> ```
> 
> I believe `V` should be `V_flow`.

Agreed, for the units to work out that variable has to be `m^3/s`. I corrected it. GLHEDT is now more in line with GLHEPRO. 

Validation



> > I'm not saying what we have done here for Coaxial is _wrong_. I'd just like to know if based off the physics we think its _right_.
> 
> It is still possible that there is an issue with `Coaxial`. Any difference with the other tools should be looked at.

I agree.

MassimoCimmino:
I am still a bit puzzled.

`R_b` seems to be correctly calculated by the `multipole` function:
```
>>> R_fg = R_fp + gt.pipes.conduction_thermal_resistance_circular_pipe(r_out_out, r_b, k_g)
>>> print(R_fg, Coaxial._Rd[1,1])
0.17950553744780107 0.17950553744780104
```

Using `eta` also compares well with the `borehole_thermal_resistance` function : 
```
>>> Rb_coaxial = gt.pipes.borehole_thermal_resistance(Coaxial, m_flow_borehole, fluid.cp)
>>> eta = H / (m_flow_borehole * fluid.cp) * 0.5 / Coaxial._Rd[1,1] * np.sqrt(1 + 4.*Coaxial._Rd[1,1] / R_ff)
>>> print(eta, Coaxial._Rd[1,1], eta / np.tanh(eta) * Coaxial._Rd[1,1], Rb_coaxial)
2.5851751131541634 0.17950553744780104 0.46935742731586894 0.46935742731586944
```

pygfunction/examples/custom_borehole.py
```diff
@@ -124,7 +129,25 @@ def main():
...
+    Ra = R_conv_1 + R_cyl_1 + R_conv_2
+    Rb = R_conv_3 + R_cyl_2 + R_grout
+    R_12 = Ra
+    eta = H / (fluid.rho * fluid.cp * V_flow_borehole) * 1 / (2 * Rb) * \
+          np.sqrt(1 + 4 * Rb / R_12)  # eq. (3.69)
+    Rb_star_UBH = Rb * eta * (1 / np.tanh(eta))  # eq. (3.68)
+
+    Rb_ = Coaxial._Rd[1, 1]  # Local borehole thermal resistance
+    assert Rb_ == Rb
+    assert R_ff == Ra
```
j-c-cook, line 145: Yes, I see what you are saying. The assertions are true here.

j-c-cook: However, running the script comes up with 

```
Coaxial tube Borehole thermal resistance: 0.1789 m.K/W
Effective borehole resistance m.K/W(GLHEDT): 0.126910
```

Therefore, equation 3.69 and 3.68 are computing a different effective borehole thermal resistance. Why?

j-c-cook:
Strangely, replacing Coaxial._Rd[iInner, iInner] with the effective borehole resistance and then recomputing provides the effective borehole resistance found by GLHEDT and GLHEPRO. 

```
    Coaxial._Rd[0, 0] = R_b
    Coaxial._initialize_stored_coefficients()
    R_b_recalculated = gt.pipes.borehole_thermal_resistance(Coaxial, m_flow_borehole, fluid.cp)
    print('Coaxial tube resistance recalculated: {}'.format(R_b_recalculated))
```
```
Coaxial tube Borehole thermal resistance: 0.1789 m.K/W
Effective borehole resistance m.K/W(GLHEDT): 0.126910
Coaxial tube resistance recalculated: 0.12691099266737843
```

pygfunction/examples/custom_borehole.py
```diff
@@ -141,13 +142,29 @@ def main():
...
     assert Rb_ == Rb
     assert R_ff == Ra
 
+    print(R_conv_1)
+
+    def func(_Rd_i):
+        Coaxial._Rd[0, 0] = _Rd_i
+        Coaxial._initialize_stored_coefficients()
+        Rb_calc = gt.pipes.borehole_thermal_resistance(Coaxial, m_flow_borehole, fluid.cp)
+        return Rb_star_UBH - Rb_calc
```
j-c-cook, line 155: I decided that `np.inf` for that value looked a bit peculiar. So I found the value of `Coaxial._Rd[0, 0]` that drove the `func`s value to its root.

j-c-cook: Nevermind, that's a coincidence.

pygfunction/examples/custom_borehole.py
MassimoCimmino: Something strange here is that we should not find 'R_b^star < R_b (R_b = R_fg)'. Is `R_fg` the same in both tools?

j-c-cook:
This line `Rb_star_UBH = Rb * eta * (1 / np.tanh(eta))  # eq. (3.68)` hardly does anything (< 1.0e-06 difference) over a wide range of flow rates.

pygfunction/examples/custom_borehole.py
j-c-cook: This line `Rb_star_UBH = Rb * eta * (1 / np.tanh(eta))  # eq. (3.68)` hardly does anything (< 1.0e-06 difference) over a wide range of flow rates.

j-c-cook: > Something strange here is that we should not find `R_b^star < R_b (R_b = R_fg)`. Is `R_fg` the same in both tools?

`R_fg` is the local borehole thermal resistance right? The answer is yes, they are identical. The line `assert Rb_ == Rb` has to be updated to `assert abs(Rb_ - Rb) < 1.0e-6`. They are identical.

j-c-cook: > This line `Rb_star_UBH = Rb * eta * (1 / np.tanh(eta)) # eq. (3.68)` hardly does anything (< 1.0e-06 difference) over a wide range of flow rates.

I believe it could be possible that there is something wrong here. This is practically local borehole resistance. Hellstrom's thesis describes the delta circuit for annular dicts on pages 79-80. 

`pygfunction` is finding a different value for effective borehole resistance (`Rb*`) than local borehole resistanace, the GLHEDT method is not. There is going to be short circuiting in a concentric annulus. I'd think there would be more short circuiting in a concentric tube heat exchanger than a U-tube arrangement where the tubes are separated by grout.

j-c-cook:
> > I'm not saying what we have done here for Coaxial is _wrong_. I'd just like to know if based off the physics we think its _right_.
> 
> It is still possible that there is an issue with `Coaxial`. Any difference with the other tools should be looked at.

The reference for GLHEDT is Grundman (2016) pages 69-72. It is attached. I'm questioning the fact that Grundman does not mention the delta circuit, but that Hellstrom (1991) describes on pages 79-80 that the delta circuit applies to the resistance network for annular ducts. (https://github.com/MassimoCimmino/pygfunction/pull/79#discussion_r695139058)

[Grundmann_2016_MS_Thesis.pdf](https://github.com/MassimoCimmino/pygfunction/files/7041690/Grundmann_2016_MS_Thesis.pdf)

pygfunction/pipes.py
```diff
@@ -2103,7 +2252,7 @@ def conduction_thermal_resistance_circular_pipe(r_in, r_out, k_f):
     --------
 
     """
-    R_p = np.log(r_out/r_in)/(2*pi*k_f)
+    R_p = np.log(r_out/r_in)/(2*pi*k_p)
```
j-c-cook, line 2255: This was a typo that I fixed.

MassimoCimmino:
@j-c-cook I started the review on [my branch](https://github.com/MassimoCimmino/pygfunction/tree/issue36_CoaxialPipes). I will let you know if anything comes up.

Remaining checks :
- [ ] `convective_heat_transfer_coefficient_concentric_annulus()`
- [ ] `custom_borehole.py`
- [ ] documentation
- [ ] cross-validation with the other tools

PR #127, opened 2021-07-14
issue124 Enhance Documentation for variables that are used in several places

### LINKED ISSUE #124: Enhance Documentation for variables that are used in several places
When `m_flow` is requested, is this the system mass flow rate or the mass flow rate per borehole desired?

I know that the variable `m_flow` is used in many places and maybe we don't want as detailed a description there, but what would you think about having a  `nomenclature` page that describes the variables used in _pygfunction_ and their associated meaning?

Maybe a table similar to [this](http://www.coolprop.org/coolprop/HighLevelAPI.html#table-of-string-inputs-to-propssi-function)?

MassimoCimmino:
This closes #124.


MassimoCimmino:
@j-c-cook Could you take a look at this PR and confirm that it addresses the issues raised in #124?

j-c-cook:
What you have done is great. I also had a couple of other things I had thought about addressing with that issue. Please see the following PR: https://github.com/MassimoCimmino/pygfunction/pull/128#issue-690243368.

I thought it might be nice to separate the examples and modules into their own folders. What do you think?

Also, what do you think about the nomenclature table? Perhaps we should have more than one. I don't think this should be all inclusive, but rather the thermodynamic and geometric variables used.

MassimoCimmino:
I just updated the code following the merge of #128.

PR #128, opened 2021-07-14
Issue124 enhance variable documentation

j-c-cook:
There are two main purposes of this PR:

1. Move the module and example documentation files into folders respectively labeled modules and examples
2. Create a new section by the name of `nomenclature` that has a table of all of the variables uses, their units and a brief description


MassimoCimmino:
This looks great, moving the documentation for the examples and modules into separate folders is a really good idea.

I will make a careful review very soon. The nomenclature tables look good too. I may add some text after merging.

I found a file that should not be in the PR, see my review comment.

doc/source/nomenclature_tables/.~lock.heat_transfer.csv#
```diff
@@ -0,0 +1 @@
+,jackcook,beowulf,14.07.2021 17:31,file:///home/jackcook/.config/libreoffice/4;
```
MassimoCimmino, line 1: I don't think this file is required.

j-c-cook:
Thank you! I'm glad you like it. I got rid of that unnecessary file. 

On the nomenclature table, its not all inclusive because there are many arguments such as `nSegments` that don't have units. My main purpose was to include variables that have units. I thought having it all in one location may be a helpful reference to someone who doesn't know `pygfunction` as well.

MassimoCimmino:
I understand and agree not to make them all inclusive. Your tables makes me realize that there is some discrepancy in the variable naming conventions (e.g. `Rb` vs `R_pipe` -- notice the underscore) and also conflicts between different modules (e.g. `Q` is the total heat transfer rate (in Watts) in the `pipes` module but the linear heat transfer rate (in Watts per meter) ine the `load_aggregation` module).

I will merge and try to address these issues on my branch.

-------------------
@all-contributors please add @j-c-cook for doc.

Issue #133, opened 2021-08-09
Enhance nSegment definition to be int or list


j-c-cook:
In every current description of `boreSegment`s in _pygfunction_, the number of segments is always declared as an integer and a uniform segment length is used across all segments. However, what if a user wants to define a number of segments on a per borehole basis? This issue seeks to address the current inability to define number of segments on a per borehole basis.

MassimoCimmino:
Solved by #137.

PR #134, opened 2021-08-09
Enhance segment definition

j-c-cook:
This is an enhancement to the definition of the variable `nSegments` to now allow a list in addition to a positive integer to be defined. If `nSegments` is defined as a list, then the length of the list must be equivalent to the number of boreholes and each borehole must have a segment length greater than or equal to 1. 

related to #133


j-c-cook:
I plan on adding example file and documentation in PR #136. If you want a script to test the functions, see the example file in #136. 

@MassimoCimmino

MassimoCimmino:
This looks great. I will review when I can.

I would have expected changes to be needed in [`_find_axial_borehole_pairs()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L2036) to work properly with unequal number of segments.

At a glance, some changes are most probably needed to [`_heat_extraction_rates()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L464), [`_heat_extraction_rate_profiles()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L495), [`_temperatures()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L558) and [`_temperature_profiles()` ](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L589) to support the feature for visualization options.

_Note to self_ : Check if this is or can be supported by the `network` module.

j-c-cook:
> I would have expected changes to be needed in [`_find_axial_borehole_pairs()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L2036) to work properly with unequal number of segments.

I’ll need to take another look at this. I didn’t  see any dependency on ‘nSegments’ in this function. It currently runs as if there are no issues (ie. the results match in the example file). 

> At a glance, some changes are most probably needed to [`_heat_extraction_rates()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L464), [`_heat_extraction_rate_profiles()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L495), [`_temperatures()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L558) and [`_temperature_profiles()` ](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L589) to support the feature for visualization options.
 
I hadn’t noticed these because they aren’t made use of to determine the g-function. I’d refer to these as auxiliary functions. They are useful for visually understanding what is going on via plotting, but do not have an impact on the accuracy of the calculation.

I should be able to get around to this tonight or tomorrow morning.

MassimoCimmino:
> I’ll need to take another look at this. I didn’t see any dependency on ‘nSegments’ in this function. It currently runs as if there are no issues (ie. the results match in the example file).

The issue (if there is one) would appear here:
https://github.com/j-c-cook/pygfunction/blob/62037cd531bb2121cb190e651f68b4363468fb07/pygfunction/gfunction.py#L1853-L1863

Axial segment pairs are only mapped for the first pair of boreholes in `self.borehole_to_borehole[n]`. Adding the number of segments as a criteria in `_find_axial_borehole_pairs()` would ensure that different pairs `(nq_i, nq_j)` are treated separately.

j-c-cook:
> > I would have expected changes to be needed in [`_find_axial_borehole_pairs()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L2036) to work properly with unequal number of segments.
> 
> I’ll need to take another look at this. I didn’t see any dependency on ‘nSegments’ in this function. It currently runs as if there are no issues (ie. the results match in the example file).

I think you're saying that a uniform segments lengths assumption is somehow baked in even with my modification to _map​_axial_segment_pairs() to accept (i, j) rather than (boreholes[i], boreholes[j]). I don't think I quite understand the code well enough to see what you're pointing out. 

https://github.com/j-c-cook/pygfunction/blob/94a1eac2e8fbbc9d38a520aaa3d85c20b0fbf949/pygfunction/gfunction.py#L1867-L1868

> > At a glance, some changes are most probably needed to [`_heat_extraction_rates()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L464), [`_heat_extraction_rate_profiles()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L495), [`_temperatures()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L558) and [`_temperature_profiles()` ](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L589) to support the feature for visualization options.
> 
> I hadn’t noticed these because they aren’t made use of to determine the g-function. I’d refer to these as auxiliary functions. They are useful for visually understanding what is going on via plotting, but do not have an impact on the accuracy of the calculation.

Done.

tests/gfunction_test.py
```diff
@@ -162,6 +162,42 @@ def test_three_by_two_twelve_segments(self, rel_tol=1.0e-4):
...
+        g_ref = gfunc.gFunc
+
+        # g-Function calculation with nSegments passed as list, where the list
+        # is of the same length as boreField and each borehole is defined to
+        # have >= 1 segment
+        # Note: nSegments[i] pertains to boreField[i]
+        options = {'nSegments': [nSegments] * len(boreField)}
+        gfunc = gFunction(boreField, self.alpha, time=time, options=options)
+
+        self.assertEqual(g_ref, gfunc.gFunc)
```
j-c-cook, line 196: I added in a function for testing unequal segments.

MassimoCimmino:
Thanks @j-c-cook.

I will probably update the example and test to have a list of unequal numbers of segments after I work on `_find_axial_borehole_pairs()`.  With the way you approached the implementation, it should not be to complicated.

I have been unable to run a case with unequal numbers of segment, for example:
```
    # Simulation parameters

    # Borehole dimensions
    D = 4.0             # Borehole buried depth (m)
    H = 150.0           # Borehole length (m)
    r_b = 0.075         # Borehole radius (m)
    B = 7.5             # Borehole spacing (m)

    # Thermal properties
    alpha = 1.0e-6      # Ground thermal diffusivity (m2/s)

    # Geometrically expanding time vector.
    dt = 100*3600.                  # Time step
    tmax = 3000. * 8760. * 3600.    # Maximum time
    Nt = 50                         # Number of time steps
    ts = H**2/(9.*alpha)            # Bore field characteristic time
    time = gt.utilities.time_geometric(dt, tmax, Nt)

    # Borehole field

    # Field of 2x2 (n=4) boreholes
    N_1 = 2
    N_2 = 2
    boreField = gt.boreholes.rectangle_field(N_1, N_2, B, B, H, D, r_b)

    # Evaluate g-functions with different segment options

    nSegments = [8, 12, 16, 20]  # number of segments toggle
    options = {'nSegments': nSegments, 'disp': True,
               'profiles': True}
    gfunc = gt.gfunction.gFunction(
        boreField, alpha, time=time, options=options)

    print(gfunc.gFunc)

    gfunc.visualize_heat_extraction_rates()
    gfunc.visualize_heat_extraction_rate_profiles()
    gfunc.visualize_temperatures()
    gfunc.visualize_temperature_profiles()
```

pygfunction/gfunction.py
```diff
@@ -2344,11 +2344,13 @@ def _map_segment_pairs(self, i_pair, j_pair, k_pair, borehole_to_borehole,
...
+                j_sum = j * nq_j
             else:
-                nq_i = self.nSegments[i]  # number of sources in borehole i
-                nq_j = self.nSegments[j]  # number of sources in borehole j
-            i_segment.append(i_pair + i * nq_i)
-            j_segment.append(j_pair + j * nq_j)
+                i_sum = sum(self.nSegments[0:i])
+                j_sum = sum(self.nSegments[0:j])
+            i_segment.append(i_pair + i_sum)
+            j_segment.append(j_pair + j_sum)
```
j-c-cook: With the example you give, I was getting an out of bounds error with the `h_ij` matrix. I had made a mistake, its now corrected.

j-c-cook: However, I am now getting a Singular matrix error. 

```
  File "/media/jackcook/SHARED/Masters/Research/pygfunction/pygfunction/gfunction.py", line 1303, in solve
    X = np.linalg.solve(A, B)
  File "<__array_function__ internals>", line 6, in solve
  File "/home/jackcook/anaconda3/envs/GLHE/lib/python3.6/site-packages/numpy/linalg/linalg.py", line 394, in solve
    r = gufunc(a, b, signature=signature, extobj=extobj)
  File "/home/jackcook/anaconda3/envs/GLHE/lib/python3.6/site-packages/numpy/linalg/linalg.py", line 88, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")
numpy.linalg.LinAlgError: Singular matrix
```

Not sure what the problem is at the moment.

j-c-cook: There appears to be some same-segment to same-segment responses missing in `h_dt`.

pygfunction/gfunction.py
```diff
@@ -1681,16 +1691,28 @@ def thermal_response_factors(self, time, alpha, kind='linear'):
...
+                        # sum all previous segments up to borehole i
+                        j0 = sum(self.nSegments[0:j])
+                        # add the number of segments in borehole i
+                        j1 = j0 + nq_j
                     h_ij[i0:i1, j0:j1, 1:] = h[:, j0-i1:j1-i1, :]
                     if j > i:
+                        # TODO: Resolve segment length scaling for reciprocal response
                         H_ratio = self.boreholes[i].H/self.boreholes[j].H
                         h_ij[j0:j1, i0:i1, 1:] = np.transpose(
                             h[:, j0-i1:j1-i1, :]*H_ratio, (1, 0, 2))
```
j-c-cook, line 1718: I have nearly added in the ability to compute g-functions when `nSegments` is off and the `method` is detailed. However, one thing I'm not quite understanding is how the reciprocity relationship could have ever been scaled by H[i]/H[j], when that is the ratio of borehole height to borehole height, _not_ a vector of segment heights to segment heights.

MassimoCimmino: Wouldn't this work (?) : `H_ratio = (self.boreholes[i].H / nq_i) / (self.boreholes[j].H / nq_j)`

j-c-cook: Agreed. Your idea works for this branch, I referred to it as a short-term resolution 4357392eb5ed93a07675c5c23a5e89cd603d0989 

I plan to move to #137 now. Let me know if you'd rather have #137 merge into #133 or #135. If not I'll open a PR for #137 after I make it happen.

MassimoCimmino: I don't see how #137 is a separate issue from either of #133 and #135 -- it is only an issue in the context of these issues. Changes to the reciprocal responses are within the scope of #133 and #135. I would simply close #137. We can still make PRs to #133 and #135.

j-c-cook: I believe I have properly addressed this with b61c2879de6be2411a523ec70c9689a9050b57e5. The solution maintains speed via broadcasting. There is going to be slightly more memory consumed because I load `segment_lengths` into memory, however in #135 I plan to include a new instance variable named `segmentLengths` that stores the data long term.

pygfunction/gfunction.py
```diff
@@ -1714,7 +1714,7 @@ def thermal_response_factors(self, time, alpha, kind='linear'):
                         j1 = j0 + nq_j
                     h_ij[i0:i1, j0:j1, 1:] = h[:, j0-i1:j1-i1, :]
                     if j > i:
-                        H_ratio = segment_lengths[:, i0:i1] / \
+                        H_ratio = segment_lengths[:, i0:i0+(j1-j0)] / \
```
j-c-cook: Both of these have to do with `method: detailed`:

- b61c2879de6be2411a523ec70c9689a9050b57e5 made it possible to define different heights by making the reciprocal calculation occur via a vector of segment height ratios
- 0a4ac88ad4ca129648ccd0e822d74e1a7d4f283f (above) makes the solution run when the number of segments on each borehole is not the same

When the number of segments of `borehole1` and `borehole2` are not the same, the segment slice lengths are unequal. I made the program run by ensuring the slice lengths are the same, but some of the ratios then become 1. I'm not sure if this solution is accurate.

MassimoCimmino:
@j-c-cook see my proposed changes in [5c18b76](https://github.com/MassimoCimmino/pygfunction/commit/5c18b7673c83b0e3eaaac3c576fea09f3e62d45d). I think this solves the issues with the reciprocal factors and array lengths. It could also prepare for #135.

I will work on the review soon after validating there is no conflict with the `'MIFT` boundary condition and the rest of the modules.

j-c-cook:
The changes of 5c18b7673c83b0e3eaaac3c576fea09f3e62d45d have been pulled into this PR. Your example with varied numbers of segments runs runs for both `similarities` and `detailed`. The g-functions computed for both methods have the same solution out to 8 decimal places.

j-c-cook:
UBHWT timing of `pygfunction-2.0` before and after PR 134 would be incorporated. They are nearly identical with lower number of segments, and the differences could be in the natural load of my computer. Its not quite clear if the timing difference is negligible, or if `pyg-2.0+PR134` has incorporated an unknown (to me) time complexity. I'm going to look into this a bit closer. 

UBHWT-timing

pygfunction/gfunction.py
```diff
@@ -2067,7 +2085,8 @@ def _find_axial_borehole_pairs(self, boreholes):
                     m = borehole_to_self[k][0]
                     # Add the borehole to the group if a similar borehole is
                     # found
-                    if self._compare_boreholes(boreholes[i], boreholes[m]):
+                    if (self._compare_boreholes(boreholes[i], boreholes[m]) and
+                        self.nBoreSegments[i] == self.nBoreSegments[m]):
```
j-c-cook, line 2089: It appears that _find_axial_borehole_pairs was not modified, but rather an addition to the _compare_borehole statement.

j-c-cook:
I reduced the background processes during this timing by closing unnecessary applications. Each of the timing points shown in the plot below is an average of 5 recordings. These timings are done on a 12 x Intel Core i7-9850H CPU @ 2.60 GHz. There is negligible difference in the branches. This agrees with the modification to the code. There should be the same amount of processing occurring for a borefield with equal heights and uniform segment lengths. 

UBHWT-timing

pygfunction/gfunction.py
```diff
@@ -1296,7 +1306,7 @@ def solve(self, time, alpha):
                     # Energy conservation: sum([Q_b*H_b]) = sum([H_b])
                     # ---------------------------------------------------------
                     a_in, a_b = self.network.coefficients_borehole_heat_extraction_rate(
-                            self.network.m_flow_network, self.network.cp_f, self.nSegments)
+                            self.network.m_flow_network, self.network.cp_f, self.nBoreSegments)
```
MassimoCimmino, line 1309: > I will work on the review soon after validating there is no conflict with the `'MIFT` boundary condition and the rest of the modules.

It turns out I already previously implemented support for unequal numbers of segments in the `Network` object. This seems to be the only change required.

MassimoCimmino:
@j-c-cook I just pushed my changes to your branch.

I also do not find any significant timing differences with the `master` branch. Below are single run timings from my pc.

Figure_1

This should be ready to merge. This closes #133.

j-c-cook:
Nice, I was thinking it would be nice to wrap this up today, but you beat me to it! Thank you. 

One thing about the tests -- would it be good to have a test for UIFT and mixed UIFT (network of boreholes) in the test too?

MassimoCimmino:
Yes, we could definitely add a test for networks of boreholes.

j-c-cook:
I've taken a pass at the differences. As far as I can tell, everything looks great. 

I also made some accuracy comparisons to make sure that we haven't changed the integrity of the calculations. The `np.allclose()` function returns true for checking the g-function values in the plot below.  

accuracy_PR134

MassimoCimmino:
Thanks @j-c-cook cook. I'll do one last check and then merge probably today.

MassimoCimmino (approved):

Issue #135, opened 2021-08-09
NewFeature unequal segment lengths


j-c-cook:
Each borehole should have a certain number of segments to be split into. Each segment should have a length. 

Issue #133 allows for boreholes to have unequal number of segments per borehole. Additionally, each segment length should be definable from a high level API. This issue seeks to address the definition of segment lengths. 

This new feature would allow for `SegmentLengths` of each segment to be defined in addition to the number of segments at each borehole being different.

MassimoCimmino:
I like the idea.

It might be better to consider a `lengthFraction` (list of (nq_i,) arrays) so that the sum of each array is 1, i.e. `np.abs(1 - np.sum(x)) < tol` (using the same length tolerance as currently implemented in the `gFunction` object). In any case, it can be regularized with `x = x/np.sum(x)` when formatting inputs.

Here, I foresee a use case where the same bore field configuration is calculated with different lengths. Having fractions as inputs avoids re-evaluating the length distributions.

A nice feature would be to add some functions in the `utilities` module for some recommended length distributions (to be determined).

As I mentioned in the #134 thread, it might be required to modify [`_find_axial_borehole_pairs()`](https://github.com/MassimoCimmino/pygfunction/blob/369aa2d425c517ffdde86621144e97950034d8f2/pygfunction/gfunction.py#L2036). "Similar" borehole pairs are determined based on `H`, `D`, `r_b`, `nq_i` and now the length fractions.

j-c-cook:
> A nice feature would be to add some functions in the `utilities` module for some recommended length distributions (to be determined).

Glad to hear you would like to see that. It is a current area of research. I think that should be apart of a future PR.

MassimoCimmino:
I just realized there might be some difficulties to make this compatible with the `Pipe` and `Network` objects. I will try to figure something out while working on #132.

j-c-cook:
The [pull request](https://github.com/MassimoCimmino/pygfunction/pull/136) is [merged](https://github.com/MassimoCimmino/pygfunction/pull/136#event-5551061452) .

j-c-cook:
Can you link this issue to [Milestone v2.1](https://github.com/MassimoCimmino/pygfunction/milestone/6)?

Issue #132, opened 2021-07-16
Refactor for loops in `Pipe` objects


MassimoCimmino:
`Pipe` objects contain for loops for the evaluation of coefficients in a few class methods, e.g. in `SingleUTube._general_solution()`, `MultipleUTube._continuity_condition()`, etc. These for loops are computationally expensive, especially when called by `Network` objects during simulations.

These loops could be refactored to use array operations, as was done in #119 for `Network._network_coefficients_from_pipe_coefficients()`.

MassimoCimmino:
## Test case for single U-Tube

```
import matplotlib.pyplot as plt
import numpy as np
from scipy.constants import pi

import pygfunction as gt

# Borehole dimensions
D = 4.0             # Borehole buried depth (m)
H = 150.0           # Borehole length (m)
r_b = 0.075         # Borehole radius (m)

# Bore field geometry (rectangular array)
N_1 = 6             # Number of boreholes in the x-direction (columns)
N_2 = 4             # Number of boreholes in the y-direction (rows)
B = 7.5             # Borehole spacing, in both directions (m)

# The field is a retangular array
boreField = gt.boreholes.rectangle_field(N_1, N_2, B, B, H, D, r_b)

# Pipe dimensions
rp_out = 0.0211     # Pipe outer radius (m)
rp_in = 0.0147      # Pipe inner radius (m)
D_s = 0.052         # Shank spacing (m)
epsilon = 1.0e-6    # Pipe roughness (m)

# Pipe positions
# Single U-tube [(x_in1, y_in1), (x_out1, y_out1)]
pos = [(-D_s, 0.), (0., D_s)]

# Ground properties
k_s = 2.0           # Ground thermal conductivity (W/m.K)
T_g = 10.0          # Undisturbed ground temperature (degC)

# Grout properties
k_g = 1.0           # Grout thermal conductivity (W/m.K)

# Pipe properties
k_p = 0.4           # Pipe thermal conductivity (W/m.K)

# Fluid properties
m_flow_borehole = 0.25      # Total fluid mass flow rate per borehole (kg/s)
m_flow = m_flow_borehole*N_1*N_2    # Total fluid mass flow rate (kg/s)
# The fluid is propylene-glycol (20 %) at 20 degC
fluid = gt.media.Fluid('MPG', 20.)
cp_f = fluid.cp     # Fluid specific isobaric heat capacity (J/kg.K)
den_f = fluid.rho   # Fluid density (kg/m3)
[...]

PR #136, opened 2021-08-09
NewFeature unequal segment lengths

j-c-cook:
This builds on #133 and contains the commits from PR #134. This is related to #135 and should not be merged until after #134 is. 

The segments function in boreholes.py had previously only accounted for the segment lengths to be equal on a specifi borehole. This modification provides the ability for the segments to have different lengths.

The gfunction.py file needed to address unequal segment lengths in the _map_axial_segment_pairs function so that the oreholes.segments function could properly create bore segments. The _baseSolver object had updates to the segment_lengths(), borehole_segments() and _check_inputs() functions.

  - The summation of the segment lengths for each borehole are asserted to be within a tenth of a meter of the borehole height
  - The borehole_segments() function accesses the new variable segmentLengths (if its not None) to split boreholes into segments
  - The segment_lengths() function flattens the 2D segmentLengths list if segmentLengths is not None


j-c-cook:
@MassimoCimmino I think this is ready to go. Can you take a look?

MassimoCimmino:
I see potentially similar issues as in #133 / #134:

- The distribution of segment lengths should be compared when similarities are identified (like the number of segments are now checked);
- This might affect the visualization methods of the `gFunction` class.

I would suggest a list of arrays of `segmentLengthRatios` instead of `segmentLengths` to avoid the user have to regenerate the list any time the borehole length changes. I would expect any recommendation on the "optimal" discretization scheme to be in the form of ratios rather than lengths.

I will move on to this PR after #132, #138 and #146.

Also to be considered here is that the functionality for unequal segment lengths needs to be added to the `pipe` and `network` classes to preserve the full functionality of pygfunction. Depending on how complex the implementation is, we can consider the PR for `v2.1` or `v2.2`.

MassimoCimmino:
I will be working on this PR on [my branch](https://github.com/MassimoCimmino/pygfunction/tree/issue135_NewFeatureUnequalSegmentLengths).

I have implemented segment length ratios, updated visualization methods, and fix a few issues related to `'similarities'` and the construction of the thermal response factor matrices. I will now be moving on to the `pipe` and `network` classes.

j-c-cook:
Okay sounds good, I'm glad you're open for this being in `v2.1`, but I'll understand if it gets pushed back to `v2.2`. How is that looking? I've been busy with GLHEDT and it looks like I will be for the next couple months as I try to push out a report on the subject.

It looks like you are working hard and doing some rapid development. The progress is exciting. Thank you.

MassimoCimmino:
I have also been busy with the start of the Fall semester. I have been looking into the `pipes` and `networks` module and it doesn't look like it should be too difficult to implement the variable segment lengths.

I am excited with the results so far with the `'UBWT'` condition. It seems unequal segment lengths can get more accurate results faster and with fewer segments. I am looking forward to have this in `v2.1`. It should also be compatible with the recent `'equivalent'` solver from #146.

j-c-cook:
I removed the conflicts and merged master in.

MassimoCimmino:
With the changes in [eef27e2](https://github.com/MassimoCimmino/pygfunction/pull/136/commits/eef27e2391b87d74e9b229aed890ac35739d43e6), the feature should be functional with the `'detailed'` and `'similarities'` solvers and all boundary conditions.

Some remaining tasks:

- [ ] Update the `'equivalent'` solver, and the `_EquivalentBorehole` and `_EquivalentNetwork` classes following the merge from master.
- [ ] Complete docstrings.
- [ ] Complete input checks and formatting in all classes.
- [ ] Validate the results and benchmark the accuracy/speed. A reasonable discretization is the one proposed by Eskilson (1987) with segment lengths increasing with a factor `sqrt(2)` towards the center of the boreholes.
- [ ] Complete documentation.
- [ ] Complete examples.
- [ ] Complete tests.
- [ ] Possible : Review how boreholes are compared in the `gFunction` classes with the recent inclusion of varying `nSegments` and `segment_ratios` (e.g. [here](https://github.com/MassimoCimmino/pygfunction/pull/136/files?file-filters%5B%5D=.py#diff-6869c048747b96f8229c34526a042448c5a68b881bf150c637044c0f3b97607cR2232-R2245)). These checks should at the very least be reordered to have the fastest checks first (i.e. the number of segments). Also consider keeping `None` as a possible value for `segment_ratios` like in the `pipe` and `network` modules. This would be faster if the discretization is uniform instead of relying on `numpy.allclose()`.

pygfunction/networks.py
```diff
@@ -477,12 +477,9 @@ def coefficients_inlet_temperature(
...
             # [T_{f,b,out}] = [A]*[T_{f,b,in}] + [B]*[T_{b}]
-            AB = list(zip(*[
-                self.p[i].coefficients_outlet_temperature(
-                    self._m_flow_borehole[i],
-                    self._cp_borehole[i],
-                    self.nSegments[i],
-                    segment_ratios=self._segment_ratios[i])
+            AB = list(zip(*[self.p[i].coefficients_outlet_temperature(
+                self._m_flow_borehole[i], self._cp_borehole[i],
+                self.nSegments[i], segment_ratios=self._segment_ratios[i])
```
j-c-cook, line 482: Breaking here in the first test.

j-c-cook:
I worked docstrings, examples and tests. 

The test was breaking, and I fixed it. Now trying to compute a segment_ratios MIFT g-function in the test its breaking. I've run out of time on this for now. Will pick up here if you, @MassimoCimmino haven't by the time I get back to this.

MassimoCimmino:
I fixed some errors which produced incorrect results for the g-functions. Tests should also be working now.

j-c-cook:
I concur, tests are running. Do you agree that the following is where we stand?

> 
> Some remaining tasks:
> 
> * [ ]  Update the `'equivalent'` solver, and the `_EquivalentBorehole` and `_EquivalentNetwork` classes following the merge from master.
> * [x]  Complete docstrings.
> * [ ]  Complete input checks and formatting in all classes.
> * [ ]  Validate the results and benchmark the accuracy/speed. A reasonable discretization is the one proposed by Eskilson (1987) with segment lengths increasing with a factor `sqrt(2)` towards the center of the boreholes.
> * [x]  Complete documentation.
> * [x]  Complete examples.
> * [x]  Complete tests.
> * [ ]  Possible : Review how boreholes are compared in the `gFunction` classes with the recent inclusion of varying `nSegments` and `segment_ratios` (e.g. [here](https://github.com/MassimoCimmino/pygfunction/pull/136/files?file-filters%5B%5D=.py#diff-6869c048747b96f8229c34526a042448c5a68b881bf150c637044c0f3b97607cR2232-R2245)). These checks should at the very least be reordered to have the fastest checks first (i.e. the number of segments). Also consider keeping `None` as a possible value for `segment_ratios` like in the `pipe` and `network` modules. This would be faster if the discretization is uniform instead of relying on `numpy.allclose()`.
> * [ ]  Only use the stored `m_flow` and `cp` in the initialization of the `gFunction` class with the `MIFT` condition.

MassimoCimmino:
Yes, I agree. I will still make a careful review of the PR once all tasks are complete.

j-c-cook:
I plan to write a function that will discretize the boreholes based on Eskilson's proposed `sqrt(2)` . I'll do an analysis and see what a good starting segment length would be. I'll do it for rectangles and U's for varying B/H ratios for the `MIFT`.

The other points I'm going to have to leave to you.

MassimoCimmino:
Great. If it helps, here is a few lines from a script I was working on to test the convergence of equal and unequal segments:
```
        dz_min = 0.5 * (np.sqrt(2) - 1) / (np.sqrt(2)**int(n/2) - 1)
        dz = np.array([dz_min*np.sqrt(2)**(n) for n in range(int(n/2))])
        segment_ratios = np.concatenate((dz, dz[::-1]))
```

j-c-cook:
The current discretization function I am using is given below. The function gives a starting segment length ratio, and then adds segments that are multiplied by `sqrt(2.)`. When a segment length would go beyond the center of the borehole, the while loop is broken, the remaining segment length is found, multiplied by 2, and then appended. 
```
def discretize(H, end_length_ratio=0.05, ratio=np.sqrt(2.)):

    H_half = H / 2.
    sqrt_2 = ratio
    end_length = H * end_length_ratio

    segment_lengths = []
    nSegments = 0
    length = end_length
    while (sum(segment_lengths) + length) <= H_half:

        segment_lengths.append(length)
        length *= sqrt_2

        nSegments += 1

    diff = H_half - sum(segment_lengths)
    segment_lengths.append(diff * 2.)
    # segment_lengths.extend([diff, diff])

    segment_lengths = segment_lengths + list(reversed(segment_lengths[0:len(segment_lengths)-1]))

    nSegments = (nSegments * 2) + 1
    assert nSegments == len(segment_lengths)

    assert (sum(segment_lengths) - H) <= 1.0e-08

    segment_ratios = [segment_lengths[i] / H for i in range(len(segment_lengths))]

    assert (sum(segment_ratios) - 1.0) <= 1.0e-8

    return segment_ratios, nSegments
```
I used `scipy.optimize.minimize` to find the value of the starting segment ratio that minimizes the error between a converged EFT g-function. The value found is near 0.059. 

Reference, or converged g-functions are computed using 24 segments. Discretized g-functions are computed, both using 9 segment discretized along the length, equal accross boreholes. The result is that the RMSE is 0.106 for EFT and 0.016 for UT. Note that the starting segment length ratio was optimized for the EFT g-function, and not the UT g-function. The g-functions are  plotted below. 

discretized

Based on these results, the discretization is a more applicable procedure for the UT boundary condition.

MassimoCimmino:
I found it odd that the g-functions did not match for short time values. I just fixed an issue with 
[9aa8b55](https://github.com/MassimoCimmino/pygfunction/pull/136/commits/9aa8b551330be5226d2420f1a459fde7de896631).

Your results for the EFT might have improved.

j-c-cook:
It absolutely did. Thank you. 

discretized

j-c-cook:
That's a really great step forward. It changes where we are at on this branch. Here's an updated checklist of remaining tasks. 

> * [x]  Update the `'equivalent'` solver, and the `_EquivalentBorehole` and `_EquivalentNetwork` classes following the merge from master.
> * [ ]  Complete docstrings (specifically for the new `utilities.discretize` function.
> * [ ]  Complete input checks and formatting in all classes.
> * [x]  Validate the results and benchmark the accuracy/speed. A reasonable discretization is the one proposed by Eskilson (1987) with segment lengths increasing with a factor `sqrt(2)` towards the center of the boreholes.
> * [ ]  Complete documentation (specifically add documentation for new `discretize` example.
> * [ ]  Complete examples (finalize `discretize` example).
> * [ ]  Complete tests.
> * [ ]  Possible : Review how boreholes are compared in the `gFunction` classes with the recent inclusion of varying `nSegments` and `segment_ratios` (e.g. [here](https://github.com/MassimoCimmino/pygfunction/pull/136/files?file-filters%5B%5D=.py#diff-6869c048747b96f8229c34526a042448c5a68b881bf150c637044c0f3b97607cR2232-R2245)). These checks should at the very least be reordered to have the fastest checks first (i.e. the number of segments). Also consider keeping `None` as a possible value for `segment_ratios` like in the `pipe` and `network` modules. This would be faster if the discretization is uniform instead of relying on `numpy.allclose()`.
> * [ ]  Only use the stored `m_flow` and `cp` in the initialization of the `gFunction` class with the `MIFT` condition.

j-c-cook:
Updated our outlook in https://github.com/MassimoCimmino/pygfunction/pull/136#issuecomment-939084913. 

It would be nice to use this new discretize function in conjunction with the `equivalent` solver. Are there tests for solvers yet, and is that within the scope of this PR?

j-c-cook:
I have added solver tests to the g-function tests. I had to relax the `rtol` value from 1e-4 to 1e-2 because I am comparing a "converged" (nSegments=24) solution to the discretized solution. I have only used the uniform temperature boundary condition in these tests. 

At this point, this is all I can think to do. Ball in your court @MassimoCimmino.

MassimoCimmino:
@j-c-cook I have been reviewing your proposed `discretize` function. I am unsure if **(1)** having the number of segments as an output is the best way to proceed, and also unsure of **(2)** the treatment of the middle segments (the remainder of the discretization).

**On (1)** : From a user point of view, I would expect that increasing the number of segments leads to a more accurate solution. The impact of  the `ratio` and `end_length_ratio` parameters is not obvious.
**On (2)** : The middle segment having the remainder of the length breaks the continuity of the expanding mesh. This does not necessarily have negative consequences on the accuracy, but could cause problems if the remainder of the length is close to zero.

--------------------

## My solution(s)

### For a given minimum length

The script below finds the expansion factor to fill the borehole with user-provided `nSegments` and `min_length`. This will be called the `"Cook"` method in the figures below.

```
def expanding_segments_minlength(nSegments, min_length=0.05):
    if is_even(nSegments):
        nz = int(nSegments / 2)
        coefs = np.zeros(nz+1)
        coefs[0] = 1 - 2 * min_length
        coefs[1] = -1
        coefs[-1] = 2 * min_length
        roots = poly.Polynomial(coefs).roots()
        factor = max(np.real(roots))
        dz = [factor**i * min_length for i in range(nz)]
        segment_ratios = np.concatenate((dz, dz[::-1]))
    else:
        nz = int((nSegments - 1) / 2)
        coefs = np.zeros(nz+2)
        coefs[0] = 1 - 2 * min_length
        coefs[1] = -1
        coefs[-2] = min_length
        coefs[-1] = min_length
        roots = poly.Polynomial(coefs).roots()
        factor = max(np.real(roots))
        dz = [factor**i * min_length for i in range(nz)]
        segment_ratios = np.concatenate((dz, np.array([factor**nz]) * min_length, dz[::-1]))
    return segment_ratios
```

### For a given expansion factor

The script below finds the minimum length (at the edges of the borehole) to fill the borehole with user-provided `nSegments` and `factor`. This will be called the `"Eskilson"` method in the figures below.

```
def expanding_segments_factor(nSegments, factor=np.sqrt(2)):
    if is_even(nSegments):
        nz = int(nSegments / 2)
        minlength = 0.5 * (factor - 1) / (factor**nz - 1)
        dz = [factor**i * minlength for i in range(nz)]
        segment_ratios = np.concatenate((dz, dz[::-1]))
    else:
        nz = int((nSegments - 1) / 2)
        minlength = 1 / (2 * (factor**nz - 1) / (factor - 1) + factor**nz)
        dz = [factor**i * minlength for i in range(nz)]
        segment_ratios = np.concatenate((dz, np.array([factor**nz]) * minlength, dz[::-1]))
    return segment_ratios
```

---------------

## Comparison of the "default" parameter values proposed by Cook (`min_length=0.05`) and Eskilson (`factor=np.sqrt(2)`)

The figure below compares the RMSE of the g-function of a 10 by 10 borehole field using the `'equivalent'` solver and the `'UBWT'` boundary condition. The reference g-function is evaluated using equal segment lengths and `nSegments=128`.

**Some observations** :
- The `"Cook"` method has a lower RMSE for small number of segments. It quickly converges to the RMSE of a uniform discretization using `nSegments=20`. Given `min_length=0.05=1/20`, this suggests that the close to `0.05` optimal value was obtained due to the reference discretization of `nSegments=24` (the one used here is not the same so the results can differ a little).
- Both of the `"Cook"` and `"Eskilson"` methods provide better accuracy at greater speed at `nSegments=8` when compared to uniform discretization at `nSegments=12`.

Figure_1

---------------

## Optimal parameter values with the number of segments

For the same case, I used `scipy.optimize.minimize` to find the values of `min_length` and `factor` that minimize the RMSE for `nSegments` ranging from 5 to 20. The figure below shows how both values change with `nSegments`.

**Some observations** :
- The minimized RMSE is the same for both methods. This is expected since the discretization scheme is the same.
- The "optimal" `min_length` is in the range of 0.01 to 0.02. The "optimal" `factor` is greater than `np.sqrt(2)`, especially for fewer segments.

Figure_2

---------------

## Revised comparison for updated parameter values (arbitrarily chosen : `min_length=0.02` and `factor=2.5`)

Figure_3

Figure_4

---------------

I pushed scripts to replicate the results [here](https://github.com/MassimoCimmino/pygfunction-issues/tree/master/issues).

MassimoCimmino:
Here is a second set of comparisons using the `'equivalent'` solver. All cases are the same. Calculations are done with the `'equivalent'` solver, except for the reference results using `nSegments=64` which are generated using the `'similarities solver'`.

## Rectangular borefields
### Comparison of g-functions using 12 segments between the PR and the master branch
It is shown here that there is no significant differences in g-function calculations when compared to the `master`.

RMSE_12_segments_equivalent

### Comparison of g-functions with a high-resolution solution (`nSegments=64`) from the `master`
It is shown here that the new discretization achieves better accuracy with fewer `nSegments` when compared to a uniform discretization.

RMSE_64_segments_equivalent

### Calculation time
This better accuracy also comes with a reduced calculation time. There is an increase in calculation time using uniform segments in the PR.

calc_time_equivalent

## Field of 100 boreholes with two different lengths
Results also show better accuracy with lower calculation time in this case.

Figure_7
two_lengths_equivalent

MassimoCimmino:
Results are update to `32394b6`. This seems to solve the issues related to calculation time.

j-c-cook:
@MassimoCimmino When you say 

>  solve the issues related to calculation time

do you mean that for the same number of segments in the g-function calculation that unequal segment lengths and equal segment lengths now perform the same?

MassimoCimmino (approved):

Issue #153, opened 2021-10-20
Optimal discretization of segments along boreholes


MassimoCimmino:
#136 showed that the optimal `segment_ratios` that minimizes the error on the *g*-function is dependent on the number of segments (and potentially on the bore field configuration).

This issue is to find optimal parameters (`end_length_ratio`) for the borehole discretization. A solution would be to generate a table that gives the optimal parameter for each value of `nSegments` up to reasonable max value (e.g. `nSegments=128`). This could be accessed by setting `end_length_ratio='optimal'` as an input to `utilities.segment_ratios()`.

MassimoCimmino:
It seems the optimal discretization is dependent on the boundary condition and on the borehole spacing. Here is some initial results for large rectangular fields using the `'equivalent'` solver and `nSegments=8`.

![Figure_1](https://user-images.githubusercontent.com/23085996/141662954-f3972076-4821-4bfe-b8be-65efcdf50405.png)

jo-isiordia:
@MassimoCimmino, continuing with your analysis, the optimal discretization is dependent on the number of segments, as well as the borehole’s configuration, including the borehole’s length, spacing, and the number of boreholes within the field. An initial evaluation of the error on the g-function, with a discretization using the current 2% end-length ratio is shown in the following figure, where the maximum relative error is presented on the vertical axis and the field characteristics, which in this case is the non-dimensional coefficient B/H, are on the horizontal axis, whereas the number of boreholes in the field is color-mapped.
![Non-optimized g-function error - 8 segments](https://user-images.githubusercontent.com/117675022/203162978-29e95244-0d40-4022-aebe-bde8aee8982d.png)
Based on these initial results, an “optimal” end-length ratio (ELR) depending on the number of segments that allows the minimization of the maximum relative error found on 11,440 different rectangular field configurations is proposed, considering a BC-UBWT (boundary condition of equal wall temperature throughout the length and for each borehole) using the “equivalent” solver, varying the borefield’s characteristics as follows.

- Length of borehole (H) from 50 m. to 300 m. with a step of 10 m.
- Bore to bore spacing (B) from 3 m. to 10 m. with a step of 2 m.
- Number of boreholes from 2 by 1 to 20 by 20 in a rectangular array.
- Time from 100 h to 1,000 years with 25 non-uniform steps using time_geometric from the pygfunction utilities module.

[...]

PR #164, opened 2021-11-08
Fix segment ratio function

### LINKED ISSUE #159: End length ratio 0 < alpha < 0.5
Should the end length ratio always fall within `0 < alpha < 0.5` or only for when `nSegments==3`?

https://github.com/MassimoCimmino/pygfunction/blob/50b54afe70ac6e670b0896d0afc8561d088e2637/pygfunction/utilities.py#L55-L57

j-c-cook:
The limit `0 < end_length_ratio < 0.5` should apply to all cases.

Also, in preparation for #153, we could consider here to allow `nSegments=2` and `nSegments=1` (which would simply return `segment_ratios=numpy.array([0.5, 0.5])` and `numpy.array([1.0])`). A warning will be required in these cases to warn the user that the `end_length_ratio` parameter is overwritten.

closes #159


j-c-cook:
Ready for your review @MassimoCimmino.

MassimoCimmino:
Merged, thank you @j-c-cook.

PR #184, opened 2021-12-27
Enhance timing performance of pipe initialization

### LINKED ISSUE #183: Enhance timing performance for initialization of pipe objects
Computing a g-function using the `equivalent borehole method` appears to take longer for a multiple u-tube layout. The attached times are from examples in the soon to be open-source, Ground Heat Exchanger Design Tool,`ghedt`. I believe the timing differences are a result of the multiple U-tube being more complex to model, and not a programming error. I think it would be nice to make this known, though I'm not sure where the right place for it would be. 

image

j-c-cook:
The `ghedt` requires several re-initializations of the pipe objects during the field selection search. It was found in #183 that designs with the double U-tube take more computation time. @MassimoCimmino pointed out that the `multipole()` and `binom()` methods could be more optimized to increase the initialization speed. 

closes #183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Enhancements
 
 * [Issue 152](https://github.com/MassimoCimmino/pygfunction/issues/152) - Vectorized `coefficients_temperature` and `_general_solution` in `pipe` objects to accept depths `z` as an array. This speeds up calculations for `get_temperature` and `get_borehole_heat_extraction_rate` class methods.
-* [Issue 183](https://github.com/MassimoCimmino/pygfunction/issues/183) - The computation duration for initialization of the pipe objects is reduced by vectorization of the resistance network calculation in `multiple()`, and a `memoization` technique is implemented to reduce computation time for repeated function calls.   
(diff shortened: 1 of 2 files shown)


j-c-cook:
Do we have a way to test the differences here? When I was using this branch in `ghedt` my tests came back as failure. I looked at the values and they were the same up to 6 or 7 decimal places, but failed on an `AlmostEqualTo` check because that is looking for 8 places. I haven't had time to look into why that was the case.

MassimoCimmino:
I had to introduce a correction to avoid division by zero when I vectorized the `multipole` functions. The latest commit [(ce88268)](https://github.com/MassimoCimmino/pygfunction/pull/184/commits/ce88268e8490f277985c0bfad64cd2b3d2d39862) should bring it back to the initial results.

j-c-cook:
I may have to reduce the number of decimal places I am testing to make my test pass. Here is the difference I am seeing that is caused by memoization. 

```
Failure
Traceback (most recent call last):
  File "/Users/jackcook/Documents/ghedt/tests/test_GHE.py", line 191, in test_double_u_tube
    self.assertAlmostEqual(37.97229212228275, max_HP_EFT)
AssertionError: 37.97229212228275 != 37.97229237870451 within 7 places (2.56421756716918e-07 difference)



Ran 3 tests in 46.590s

FAILED (failures=1)



Process finished with exit code 1
```

MassimoCimmino:
These differences do look very small and I am not sure what would cause them. Note that the tolerance on the evaluation of multipoles is `1e-5`.

PR #185, opened 2021-12-28
Increase re-initialization speed of pipe objects

j-c-cook:
Massimo has done most of the work in #183 and #184. This is to try to help with finishing touches to get the changes implemented into the master branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Enhancements
 
 * [Issue 152](https://github.com/MassimoCimmino/pygfunction/issues/152) - Vectorized `coefficients_temperature` and `_general_solution` in `pipe` objects to accept depths `z` as an array. This speeds up calculations for `get_temperature` and `get_borehole_heat_extraction_rate` class methods.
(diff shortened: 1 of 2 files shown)


j-c-cook:
@MassimoCimmino 

We could work around defining the `thermal_resistances_dict` by accessing the object itself. It would require for `thermal_resistances` to be passed *self* (technically this should be *cls*). However, there appears to be a decrease in performance. I have not investigated if this is due to the heavier weight of passing self, or that the `hasattr` and `getattr` functions are slower than accessing a dictionary. 
```
HighLevel/find_near_square.py results
=====================================
* Single U-tube
---------------
Calculation time: 18.11 seconds
Height: 130.1818 meters
Number of boreholes: 156
Total Drilling: 20308.4 meters

* Double U-tube
---------------
Calculation time: 28.29 seconds
Height: 133.2133 meters
Number of boreholes: 144
Total Drilling: 19182.7 meters

* Coaxial tube
--------------
Calculation time: 18.25 seconds
Height: 132.7505 meters
Number of boreholes: 144
Total Drilling: 19116.1 meters
```

MassimoCimmino:
I am open to any ideas that would make this implementation cleaner.

Your timings suggests there is an issue with saving or accessing the previous outputs. Timings are similar to your results from [here](https://github.com/MassimoCimmino/pygfunction/issues/183#issuecomment-1001267650), before memoization.

j-c-cook:
I'd rather have the speed of your implementation. I am going to close this PR.

PR #191, opened 2022-01-15
Refactor tests using pytest

### LINKED ISSUE #179: Transition unit tests to pytest
This issue is to replace the `unittest` with `pytest`.

`pytest` has a lot of useful features, such as parametrized tests to simplify the implementation of tests.

MassimoCimmino:
Closes #179.


MassimoCimmino:
Refactored test are different from the current `master` but cover the same classes and functions.

Reference results are generated from the latest commit on #179. This way, future PRs can be compared to present results to make sure the are no unintended changes in *pygfunction* outputs.

Since reference results are generated by *pygfunction* they need to be checked and validated before merging.

MassimoCimmino:
I ran the new tests on `v2.1` and all passed except those that use the FLS approximation (due to it not being implemented).

This can now be merged.

PR #197, opened 2022-01-24
Issue50 Inclined boreholes

### LINKED ISSUE #50: Inclined boreholes
Does the most current version account for inclined boreholes? 

Given that the segment-to-segment response factor is calculated by equation 8 (Cimmino and Bernier 2014), how would the distance formula `d_{ij}` be implemented?
equation_8

For example, currently calculation of `d_{ij}` is given by the equation 11.
equation_11

Would equation 11 need to become `d_{uv}?` Given the implementation of inclined boreholes, could it now be that `x_i == x_u is false`, and `x_j == x_v is false`?

Additionally, the location of segment u,  `(x_u, y_u)` could vary slightly along the segment. Would the method to calculate the distance `d_{uv}` result in the distance from top-to-top, center-to-center, bottom-to-bottom or some other approach?

I'm assuming this implementation would occur in  <a href="https://github.com/MassimoCimmino/pygfunction/blob/982ab6c47dae4931415f1a0f9a43bbfcfb34580c/pygfunction/gfunction.py#L709">_borehole_segments</a>. There has been extensive validation of the analytical implementation of the uniform borehole wall temperature (UBWHT) boundary condition when compared to Eskilson's numerical method for vertical boreholes. Given implementation, inclined boreholes would need compared to some of Eskilson's inclined borehole g-functions. 

---------------------------------------------------------------------------------------------------------------------------------------------------------------- 
Cimmino, Massimo, and Michel Bernier. "A Semi-analytical Method to Generate G-functions for Geothermal Bore Fields." International Journal of Heat and Mass Transfer 70 (2014): 641-50. Web.

MassimoCimmino:
Closes #50.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [Issue 50](https://github.com/MassimoCimmino/pygfunction/issues/50) - Implemented inclined boreholes for the evaluation of *g*-functions. The implementation includes an approximation of the FLS solution for inclined boreholes based on the method of Cimmino (2021) (see [Issue 138](https://github.com/MassimoCimmino/pygfunction/issues/138)). The `'equivalent'` solver is not yet supported.
 * [Issue 138](https://github.com/MassimoCimmino/pygfunction/issues/138) - Implemented the approximation of the finite line source solution of Cimmino (2021). The approximation avoids the numerical evaluation of integrals. This speeds up the calculation of g-functions when enabled.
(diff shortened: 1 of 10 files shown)


MassimoCimmino:
Inclined boreholes are only implemented for the `'detailed'` solver, since there is no known `'similarities'` or `'equivalent'` methods for inclined boreholes.

The initial implementation seems to be working and compares well with Eskilson's results (I will update the issue with figures). 

The integral of the FLS solution had to be split between the real and image parts of the solution to avoid overflow flow errors in the product `erfc(a*s)*exp(b*s**2)` (see the code below). There could be a better implementation for this. It should be explored further before merging.

https://github.com/MassimoCimmino/pygfunction/blob/a847dfd70661f2dbf08e3cbb8f84f093a5b7e00c/pygfunction/heat_transfer.py#L772-L777

MassimoCimmino:
Remaining tasks:

- [ ] Complete docstrings
- [ ] Format code
- [ ] Merge #191 and implement tests
- [ ] Add an example
- [ ] Update CHANGELOG
- [ ] Check for more efficient integration approach

MassimoCimmino:
This is now ready for review.

PR #200, opened 2022-01-24
Issue198 optimize detailed solver

### LINKED ISSUE #198: Optimize `'detailed'` solver
While testing the implementation of the inclined FLS in #50 / #197, I found that the 'detailed' solver calculates the FLS between a borehole and itself in a `for` loop, so it is called once per borehole. This could be done in one call for all boreholes at the same time.

https://github.com/MassimoCimmino/pygfunction/blob/79e0198cd8e828797a6d6832f981b34288d36281/pygfunction/gfunction.py#L1907-L1924

MassimoCimmino:
Closes #198.


MassimoCimmino:
Timings and validation are in #198.

PR #237, opened 2022-11-10
Performance Improvement of erfint

tblanke:
Hi @MassimoCimmino ,

I have implemented a new erfint function which differ between values below 4 and above. To avoid calculating to often the erf and exponential function because above 4 these values are linear depending on the absolute input. This has improved the segment-to-segment response factor calculation time by up to 50 % (see pictures), at least on my machine.

PyGFunctionNew
PyGFunctionOld

I hope you find this helpful.

With best regards


MassimoCimmino:
This looks very promising. Thank you @tblanke. I will open an issue so that we can clean up the commits and quantify the impact on accuracy and computational time.

I have one question to start : How was the choice of a value of 4 decided? I believe the scipy implementation of `erf` in scipy is base on Abramowitz and Stegun, which provide approximations of `erf` with error bounds of 1.5e-7 or 3.0e-7. The choice of 4 brings us down to approximately 1e-8, which seems reasonable.

```
>>> from scipy.special import erfinv, erf
>>> 1. - erf(4.)
1.5417257914762672e-08
>>> erfinv(1. - 1e-8)
4.052237244600268
```

tblanke:
Hi @MassimoCimmino ,

Thanks.
After I have take a look at the functions erf and exp(-x^2), I have realized that they are close to 1 / 0 at around 4 -5.
So, I have implemented a pytest function to check if the error is less then 1e-8 (it is also in the pull request) and for 4 this holds. Here is a graph of the maximal error above the threshold value:
grafik
Ater this graph we can even take 3.7 if an error of 1e-8 is fine.

MassimoCimmino:
I replicated the changes in a clean branch : [issue239_approximateErfint](https://github.com/MassimoCimmino/pygfunction/tree/issue239_approximateErfint).

It seems the performance improvement is not always positive. Cases in the `compare_gfunction_solvers.py` run faster. However, some cases in `uniform_temperature.py` run slower. I will investigate deeper.

j-c-cook:
> However, some cases in `uniform_temperature.py` run slower. I will investigate deeper.

That is peculiar. I might have some time this weekend, if not resolved by then. 

> After I have take a look at the functions erf and exp(-x^2), I have realized that they are close to 1 / 0 at around 4 -5.

This is a neat idea, @tblanke. Hoping it works out.

PR #240, opened 2022-11-11
Issue231 singular matrix

### LINKED ISSUE #231: 1s gfunction evaluation: numpy.linalg.LinAlgError: Singular matrix
When I tried to evaluate a gfunction at 1 second, I got an 'numpy.linalg.LinAlgError: Singular matrix' error.
The problem is, I think, caused by getting a all-zero matrix when calculating the thermal response factors, leading to a singular matrix. Whenever I try to run the code with an evaluation at 2 seconds, the thermal response factors are small (10^-200) but non-zero ...

The code I used was
```Python
import pygfunction as gt

borefield = gt.boreholes.rectangle_field(10, 12, 6, 6, 100, 4, 0.075)
gt.gfunction.gFunction(borefield, 1.25*10**-6, 1.).gFunc
```

MassimoCimmino:
Closes #231.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [Issue 204](https://github.com/MassimoCimmino/pygfunction/issues/204) - Added support for Python 3.9 and 3.10. [CoolProp](https://www.coolprop.org/) is removed from the dependencies and replace with [SecondaryCoolantProps](https://github.com/mitchute/SecondaryCoolantProps).
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


wouterpeere:
I can confirm this solves #231 !
Thanks!

MassimoCimmino:
The current implementation may cause issues with the `gFunction.visualize_heat_extraction_rates()` and other related methods. I need to linearize the heat extraction rates and borehole wall temperatures before merging this.

PR #243, opened 2022-11-27
Add upload to pypi step in workflow

### LINKED ISSUE #230: Publish on release workflow
Did version 2.2.0 never get pushed onto [PyPi](https://pypi.org/project/pygfunction/#history)? Would you want to setup a [publish on release](https://github.com/j-c-cook/ghedt/blob/main/.github/workflows/python-publish.yml) workflow file? I remember thinking pre-releases don't trigger this workflow.

j-c-cook:
closes #230

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,33 @@
+name: Release on PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements.txt
+        python -m pip install -r requirements_dev.txt
+        python -m pip install build
+
+    - name: Build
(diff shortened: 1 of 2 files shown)


MassimoCimmino:
Thanks @j-c-cook.

Is there a way to test this without publishing a release, or do we have to test it on a release candidate version?

I probably need to configure the secrets on the project.

mitchute:
@MassimoCimmino you can configure it to push to https://test.pypi.org/ for testing purposes. You may need to modify the deployment action to include a `repository_url`, and then have separate secrets for test.pypi.org.

j-c-cook:
Without modifying the current workflow, I think you would need to publish a release candidate to test it. 

Matt seems to be proposing an alternative. Which would require `if: github.event.release && github.event.action == 'published'` to be temporarily removed for the test to run without a release being created.

j-c-cook:
> @MassimoCimmino see here for an example. Changing the url to `https://test.upload.pypi.org/legacy/` will push it to test.pypi.org if you like during your testing.

Please note that "here" is not currently a hyper-reference.

mitchute:
This link for example: https://github.com/mitchute/SecondaryCoolantProps/blob/cdd09459f1a692aaef2924125264753a36a2841d/.github/workflows/release.yml

MassimoCimmino (approved):

PR #232, opened 2022-10-18
Replace CoolProp

mitchute:
PR replaces the CoolProp fluid properties library with the lighter, zero-dependency library [SecondaryCoolantProps](https://pypi.org/project/SecondaryCoolantProps/). CoolProp has historically been the go-to library for refrigerant and coolant props, but development activity and support as of recent years have been waning, and up until recently versions >3.8 were unsupported. It's unclear if this will persist into the future, so relying on that dependency limits the usability of pygfunction and introduces uncertainty into the project.

SCP is an ultra-lightweight, zero-dependency library focused just on secondary coolants, which IMO should be enough for this application.

Issue #204, opened 2022-05-02
python 3.9+ compatibility


wouterpeere:

wouterpeere:
It is (for me) not possible to install the latest version of pyfunction on a python version higher than 3.9 due to problems with the coolprop package.

j-c-cook:
Pygfunction is tested on Python==3.7 and Python==3.8. The last time I looked into building on Python==3.9 I also had errors. Perhaps this issue should address the Python version in the [Requirements](https://github.com/MassimoCimmino/pygfunction#requirements) section of the README (Document Python version requirements).  https://github.com/MassimoCimmino/pygfunction/blob/a7586fbd4237a66427d2c219d719268d2fc8feaa/.github/workflows/test.yml#L17-L18) 

Is there a package you are using that requires you to use Python 3.9, or could you use 3.7 or 3.8?

MassimoCimmino:
Hello @wouterpeere.

CoolProp does seem to be the cause. I found that they are planning support for py39 in `v6.4.2`. There is a dev version on their [sourceforge](https://sourceforge.net/projects/coolprop/files/) which you could use in the meantime if you need py39.

I added this issue to the `v2.2` milestone. We can check if CoolProp `v6.4.2` is out once we get closer to our release.

wouterpeere:
@j-c-cook @MassimoCimmino Thank you for your responses.

At the moment, I have no specific reason for working with a version higher than 3.8, although in the future (especially with the new, faster, python 3.11), I will be moving to a higher version.

Since I am responsible for GHEtool (https://github.com/wouterpeere/GHEtool), a tool for borefield sizing, I wanted to keep this up to date with the newer versions of python. However, GHEtool uses pygfunction for the calculation of the equivalent borehole resistance of a field, and since the latest version of pygfunction is, due to CoolProp, not available at python >=3.9, we are still using pygfunction v1.1.2 in order to assure compatibility. This is not a big problem, but it would be nice if CoolProp would release a new version so that we can rely on a newer version of pygfunction.

MassimoCimmino:
I am removing this from the `v2.2.0` milestone. We can release a version `v2.2.x` as soon as Coolprop releases a new version that support Python 3.9+.

MassimoCimmino:
@all-contributors please add @mitchute for code, ideas.

PR #244, opened 2022-11-27
Add fluid name attribute to Fluid, minor corrections to Grundmann reference

mitchute:
This PR:
- add a `name` attribute to the `Fluid` class for easier accessibility to the same scp attribute.
- Corrects the Grundmann reference.


MassimoCimmino (approved):

MassimoCimmino:
This is related to #204 / #232.

PR #254, opened 2023-01-09
Create a borefield object

### LINKED ISSUE #210: New `Borefield` class
Many functions in the `heat_transfer` and `gfunction` modules necessitate to generate arrays of borehole parameters (i.e. `H`, `D`, `r_b`, `x`, `y`, `tilt`, `orientation`) from lists of boreholes. This generates multiple instances of duplicate and not very readable code.

The new `Borefield` class will replace lists of boreholes within the modules. Anytime a user provides a list of boreholes, this list will be used to generate a `Borefield` object. Alternatively, `Borefield` objects can be provided as input.

To preserve current behavior, the following should be considered (for an instance `borefield` of the `Borefield` class):

- `borefield[i]` should return the `i`-th borehole
- `borefield[i0:i1]` should return an instance of the `Borefield` class formed by boreholes `i0` through `i1-1`.

To simplify the code, the `Borefield` class will:

- Implement class methods similar to the `Borehole` class to evaluate, e.g., distance vectors and arrays. An additional parameter `outer` can be included to evaluate distance vectors (`outer=False`) and arrays (`outer=True`).
- Return arrays of parameters, e.g. `borefield.H` returns an array of borehole lengths.

j-c-cook:
closes #210

### DIFF
--- a/pygfunction/boreholes.py
+++ b/pygfunction/boreholes.py
@@ -237,6 +237,38 @@ class Borehole(object):
         return z
 
 
+class Borefield(list):
+    def __int__(self, boreholes: list = None):
+        if type(boreholes) == list:
+            super.__init__(boreholes)
+        elif type(boreholes) == Borehole:
+            super.__init__([boreholes])
+        else:
+            super.__init__()
+
+    def H(self):
+        return np.array([self.__getitem__(i).H for i, _ in enumerate(self)])
+
+    def D(self):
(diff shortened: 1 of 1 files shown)


j-c-cook:
@MassimoCimmino This is ready for some feedback when you get a chance.

j-c-cook:
Resolved in #308.

Issue #266, opened 2023-09-04
g-Function becomes negative for coaxial pipe depending on fluid 


OskarRaftegard:
I've was playing around with a coaxial pipe flows/dimensions, when the g-function suddenly became negative. 

G-function is positive for 20 degC water (I think it is a laminar flow in the inner pipe as well as the outer annulus)
It is negative for 60 degC water (I think it is laminar flow in the inner pipe and turbulent in the outer annulus)

I based my code on the examples in the documentation, but I could of course have made mistakes.

Below is my code:

```
# -*- coding: utf-8 -*-


import numpy as np
import pygfunction as gt
from scipy.constants import pi
import matplotlib.pyplot as plt
from matplotlib.ticker import AutoMinorLocator

# -------------------------------------------------------------------------
#  EXAMPLE OF SETTINGS - 
# -------------------------------------------------------------------------



B = 4               # Distance between boreholes (m)
D = 4.0             # Borehole buried depth (m)
r_b = 0.1143 / 2    # Borehole radius (m)
H = 200.            # Borehole length (m)


# ---- DETAILED SETTINGS -----

# g-Function calculation options
nSegments = 8
method = "similarities"
boundary_condition= "MIFT"

Jmultipoles = 4 # Used by pipes module

# Simulation parameters
dt = 3600.                  # Time step (s)
tmax = 100.*8760. * 3600.    # Maximum time (s)


# Load aggregation scheme
cells_per_level = 10 # 5 is defualt value
nSources=1


# Borefield geometry
N_1 = 1           # Number of boreholes in the x-direction (columns)
N_2 = 1             # Number of boreholes in the y-direction (rows)

# Ground properties
T_g = 10.0          # Undisturbed ground temperature (degC)
k_s = 3.1           # Ground thermal conductivity (W/m.K)
Cp_vol = 2160000    # Volumetric heat capacity (J/m3.K)
alpha = k_s/Cp_vol      # Ground thermal diffusivity (m2/s)


# Pipe dimensions and properties:
# pos, r_in, r_out, borehole, k_s, k_g, R_ff, R_fp, J

# Pipe properties
epsilon = 1.5e-6    # Pipe roughness (m)
k_p = 0.42           # Pipe thermal conductivity (W/m.K)

# Pipe positions
# Coaxial pipe (x, y)
pos = (0., 0.)

# Pipe dimensions (coaxial)
# Not: Outer pipe is 0.5 mm smaller than borehole and 0.001 mm thick
[...]

PR #284, opened 2024-06-17
Compatibility with numpy 2.0

wouterpeere:
The latest numpy version (2.0), released yesterday 06/16/2024, has breaking changes in comparison with v1.*
For pygfunction, this is rather minor, since only the np.cfloat should be changed to np.complex128 in the pipes module.

Since in the requirements of pygfunction, no upper boundary on the numpy version is defined, this can lead to major failures when installing pygfunction on a new venv.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - maintenance/2.2.x
-      - numpy2.0
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


MassimoCimmino:
Thank you @wouterpeere.

This closes #285. #285 and #266 should be the last two issues leading to `v2.2.3`. I am hopeful I can finish these up by the end of next week.

Issue #257, opened 2023-01-12
Code health and technical debt


ikijano:
Hi,

I have following this awesome project some time. The place where I'm working we use pre-calculated g-functions and I studied could we utilize this library for custom cases, but sadly it's not case in long time.

When I first time looked the pygfunction code I was slightly shocked because there is some quite a huge files and functions. I'm not python developer but I have used language in past, but I had hard time to follow the code and I had to give up.
For curiosity I ran code analysis using codescene.io community edition against pygfunction (forked) code base: https://codescene.io/projects/33542/jobs/790698/results
and results are interesting.

I see there is already raised issues to enhancement code quality:
- #120 
- #210 
- #211 
Issues shows similar findings what analysis tool found but there is some other aspects what should be considered: https://codescene.io/projects/33542/jobs/790698/results/warnings/code-health-degradation

I would like to help but my python knowledge is not good enough and my understanding the subject is very limited.

j-c-cook:
@ikijano Thank you for your constructive criticism. 

This package implements some serious theoretical physical models. Looking at this code base and considering it complex is natural, I believe. The complexity and sheer depth of knowledge that is located in this package is impressive. I implemented one small piece of `pygfunction` in C++ a couple years ago, and the task was no walk in the park. 

On top of the math expertise required here, the `numpy` package has been extensively utilized for computational speed. Therefore, a Python developer well versed in the native object containers could come here and need to educate himself/herself on NumPy lingo. The utilization of `numpy` here is impressive. 

If you want to dive into the implementation, I encourage you to pick one aspect at a time. Small pieces here are feasible to be understood, but all at one time with no aim would likely not result in anything meaningful (i.e. is there existence without limitation?). Perhaps there is a need for a certain kind of documentation along with this package. Please do not hesitate to create new issues with a question or to start a discussion at any time. Feedback is really helpful to help better tailor this package for the next person. 

We can always strive to have a more human readable implementation. The implementation vs. feature is a balancing act that is crucial. I say the balance is analogous to a physical foundation. 

As a response to the code analysis you presented, I've proposed #258. Moving to a Python code formatter like [black](https://github.com/psf/black) may be an improvement (though I really think the name of that package is horribly chosen). I also think we could implement some static checks in the build process using `mypy`.

ikijano:
Thank you @j-c-cook.
[...]

Issue #219, opened 2022-07-15
Groundwater advection


MassimoCimmino:
This issue is to implement the moving infinite line source solution (MFLS) for the evaluation of *g*-functions with groundwater flow. The solution is presented in [1].  An inclined moving finite line source should be possible following the implementation in #50.  One thing to keep in mind is that, under groundwater flow, the reciprocal thermal response factor `h_12 != H_2/H_1 * h_21`. Similarity rules can be inferred from the MFLS solution. The `'equivalent'` solver may be extended to consider groundwater flow. It is unknown if an approximation of the MFLS is possible.

The heat transport velocity can be added as an optional parameter to the `finite_line_source` function and the `gFunction` class, with `v_T = 0.`.

-------------------------
[1] Molina-Giraldo, N., Blum, P., Zhu, K., Bayer, P., & Fang, Z. (2011). A moving finite line source model to simulate borehole heat exchangers with groundwater advection. *International Journal of Thermal Sciences, 50*(12), 2506-2513.

wouterpeere:
There are a couple of improvements I think would be possible to make to the model, which are not explicitly in the paper:

1. Add a flow direction, so the orientation of the borefield w.r.t. the flow can be considered.
2. Add the flow only to a certain part of the borefield. Typically, you only have a flow at certain depths in the ground and not in certain aquitard layers. This could be an interesting addition, where you can set the range where there is flow.

I'm not 100% sure about the complexity of these additions, but since the borehole is segmented, I think it would be feasible.

wouterpeere:
Perhaps this is also an interesting read: https://www.sciencedirect.com/science/article/abs/pii/S0378778819325186

Issue #211, opened 2022-06-21
Refactor `heat_transfer` module


MassimoCimmino:
This follows #210.

Functions in the `heat_transfer` module have long lists of borehole parameters (e.g. `H`, `D`, `r_b`). With the implementation of the `Borefield` class, this can be simplified to only take `Borehole`objects, lists of boreholes or `Borefield` objects as input.

This is backward-compatible with regards to the `finite_line_source` function but not to the other functions in the modules.

Issue #178, opened 2021-12-14
Type annotations and mypy


MassimoCimmino:
This issue is to implement type annotations and configure `mypy` for type checking.

Type annotations were introduced in [PEP484](https://www.python.org/dev/peps/pep-0484/) for Python 3.5.

PR #310, opened 2025-01-24
Issue210 borefield class

### LINKED ISSUE #210: New `Borefield` class

MassimoCimmino:
This follows the work in #308 and works towards fully integrating the Borefield class into the g-function module and solvers to close #210 and #211.

This requires a major refactoring of the `heat_transfer` module and the different g-function solvers. Here is an overview of the ongoing work (and the reasoning) on this issue:

- [ ] **Move solvers into their own `solvers` module.** The `gfunction` module was too large and this allows to isolate each class into their own script to facilitate the work on different parts of pygfunction. This will facilitate the addition of new solvers in the future.
- [ ] **Split the `heat_transfer` module into multiple scripts.** This script was also too large. Different functions can naturally be classified into files for the vertical FLS and the inclined FLS. This change will also facilitate the addition of other heat transfer solutions in the future (e.g. for #44 and #219).
- [ ] **Refactor `heat_transfer` functions to use `Borefield` objects as inputs instead of long lists of geometrical parameters.** This will simplify unnecessary code in the solvers that formats geometrical parameters before calls to `heat_transfer` functions.
- [ ] **Refactor the `Detailed`, `Similarities` and `Equivalent` solvers to make use of the `Borefield` class.** This should greatly improve the readability of the code.
- [ ] Use type annotations in refactored modules. We already used them in the `borefield` module for #308. Since the solvers will be almost entirely re-implemented, might as well make progress on #178.
- [ ] **Update unit tests.**
- [ ] **Update documentation.**

This addresses some of the issues pointed out by @ikijano in #257. It has become increasingly difficult to implement new features as they multiply (it was not possible to predict the vast array of methods we have today when we developed v2.0) and it is time to re-visit the implementation to make future development easier.

The work on issue #210 had been delayed since it first looked like it would negatively affect the computational time. Thankfully, it now looks like there might even be some (not that significant but still not always negligible) improvements.


pygfunction/borefield.py
```diff
@@ -45,10 +47,9 @@ def __init__(
             self, H: npt.ArrayLike, D: npt.ArrayLike, r_b: npt.ArrayLike,
             x: npt.ArrayLike, y: npt.ArrayLike, tilt: npt.ArrayLike = 0.,
             orientation: npt.ArrayLike = 0.):
-        # Convert x and y coordinates to arrays
-        x = np.atleast_1d(x)
-        y = np.atleast_1d(y)
-        self.nBoreholes = np.maximum(len(x), len(y))
+        self.nBoreholes = np.max(
+            [len(np.atleast_1d(var))
+             for var in (H, D, r_b, x, y, tilt, orientation)])
```
j-c-cook, line 52: Did you consider an assertion for when `len(x) != len(y)`?

What about enforcing `H`, `D`, `r_b`, `tilt` and `orientation` to match `len(<var>) == len(y)` or `len(<var>) == 1`?

MassimoCimmino: An assertion would certainly be important here. The goal was to mimic numpy array broadcasting rules and allow for example `x = [0., 5., 10.], y = 0.` if 3 boreholes are on the same row.

j-c-cook:
The following example shows that a borehole field of size 1 can be both a `Borefield` and a `Borehole`. Have you thought about deprecating the `Borehole` object? 

```python
    x = np.array([0.])
    y = np.array([0.])
    borefield = gt.borefield.Borefield(H, D, r_b, x, y)
    print(type(borefield))
    print(type(borefield[0]))
```

Outputs:
```
<class 'pygfunction.borefield.Borefield'>
<class 'pygfunction.boreholes.Borehole'>
```

pygfunction/borefield.py
MassimoCimmino: The `Borehole` class can represent both a segment of a borehole or the borehole itself. Likewise, the collection of parameters in `Borefield` can represent segments or boreholes. With that in mind, segments along a vertical borehole share the same coordinates (`x`, `y`) and only require `H` and `D` arrays at instantiation.

One way to approach this would be to use `b = np.broadcast(H, D, r_b, x, y, tilt, orientation)` to make sure that the inputs are broadcastable to a common shape with `b.nd <= 1`. The number of boreholes is then `nBoreholes = b.size`.

j-c-cook: Feel free to make use of any ideas contained in commit https://github.com/j-c-cook/pygfunction/commit/bd6e46792064c64b76c54cf8752b839c1d249c23. I've enjoyed the back and forth and little bit of collaboration here. I'll likely drop out of discussion now.

MassimoCimmino: I adapted your changes in #322. I removed the try / except clauses since `numpy.broadcast` raises an error if the arguments cannot be broadcast.

Issue #304, opened 2024-08-30
Variable fluid convection thermal resistance in pipe objects


MassimoCimmino:
This issue will implement teh possibility to supply an array of thermal resistances `R_fp` along with an array of fluid mass flow rates to handle the variation of delta-circuit thermal resistances with regards to fluid mass flow rates. The delta-circuit thermal resistances will then be pre-computed and interpolated during simulations. Interpolation could also apply to effective thermal resistances in `Pipe` and `Network` objects to speed up simulations.

PR #315, opened 2025-04-09
Add API for MIFT Boundary Condition

mitchute:
PR adds an API for computing g-functions using static inputs based on the `MIFT` boundary condition. As a reminder, PR #308 added a parallel API for the `UHTR` and `UBWT` boundary conditions. Since the input requirements are so different from the two schemes (`MIFT` vs `UHTR` and `UBWT`), rather than trying to unify the API, IMO it looks more straightforward for users to pick which one they want based on the BC and level of inputs they are interested in providing. The RR is still somewhat "drafty," and somethings like the use of `Enum` and the right sets of input arguments to expose need to be pinned down. The code is functional for the current test cases, but we likely should be testing and exposing more, as needed to get to the appropriate basic functionality level. We could add lots of fanciness on top after that.

Feel free to review/discuss and we can go from there.

### DIFF
--- a/pygfunction/gfunction.py
+++ b/pygfunction/gfunction.py
@@ -4843,7 +4843,7 @@ def evaluate_g_function(
         reversible_flow: bool = True,
         m_flow_network: Union[float, None] = None,
         fluid_name: Union[str, None] = None,
-        fluid_concentration: Union[float, None] = None,
+        fluid_concentration_pct: Union[float, None] = None,
         nSegments: Union[int, None] = None,
         segment_ratios=None
 ):
@@ -4852,14 +4852,14 @@ def evaluate_g_function(
         options = {}
 
     borefield = Borefield(H, D, r_b, x, y, tilt, orientation)
-    nbh = len(borefield.H)
 
     if boundary_condition == "MIFT":
 
-        fluid = Fluid(fluid_name, fluid_concentration)
+        boreholes = borefield.to_boreholes()
+        fluid = Fluid(fluid_name, fluid_concentration_pct)
 
         pipes = get_pipes(
-            nbh,
+            boreholes,
             pipe_type,
             pos,
             r_in,
@@ -4874,32 +4874,21 @@ def evaluate_g_function(
             reversible_flow
         )
 
-        network = Network(
-            borefield.to_boreholes(),
-            pipes,
-        )
-
-        gfunc = gFunction(
-            network,
-            alpha,
-            time=time,
-            method=method,
-            boundary_condition=boundary_condition,
-            options=options,
-        )
+        network = Network(boreholes, pipes, m_flow_network=m_flow_network, cp_f=fluid.cp)
+        network_or_borefield = network
 
     elif boundary_condition in ["UHTR", "UBWT"]:
-
-        gfunc = gFunction(
-            borefield,
-            alpha,
-            time=time,
-            method=method,
-            boundary_condition=boundary_condition,
-            options=options,
-        )
-
+        network_or_borefield = borefield
     else:
         raise ValueError("boundary_condition must be 'MIFT', 'UHTR', or 'UBWT'")
+
+    gfunc = gFunction(
+        network_or_borefield,
+        alpha,
+        time=time,
+        method=method,
+        boundary_condition=boundary_condition,
+        options=options,
+    )
 
     return gfunc.gFunc
--- a/pygfunction/gfunction.py
+++ b/pygfunction/gfunction.py
@@ -4827,8 +4827,8 @@ def evaluate_g_function_MIFT(
         alpha: float,
         time: npt.ArrayLike,
         pos: List[tuple],
-        r_in: Union[float, tuple],
-        r_out: Union[float, tuple],
+        r_in: Union[float, tuple, list],
+        r_out: Union[float, tuple, list],
         k_s: float,
         k_g: float,
         k_p: float,
@@ -4843,7 +4843,6 @@ def evaluate_g_function_MIFT(
         orientation: npt.ArrayLike = 0.,
         reversible_flow: bool = True,
 ):
-
     if options is None:
         options = {}
 
--- a/pygfunction/gfunction.py
+++ b/pygfunction/gfunction.py
@@ -3,7 +3,6 @@ import warnings
 from time import perf_counter
 from typing import Union, Dict, List
 
-import matplotlib.pyplot as plt
 import numpy as np
 import numpy.typing as npt
 from scipy.cluster.hierarchy import cut_tree, dendrogram, linkage
@@ -19,7 +18,7 @@ from .heat_transfer import finite_line_source, finite_line_source_vectorized, \
 from .media import Fluid
 from .networks import Network, _EquivalentNetwork, network_thermal_resistance
 from .pipes import get_pipes, PipeTypes
-from .utilities import _initialize_figure, _format_axes
+from .utilities import _initialize_figure, _format_axes, segment_ratios
 
 
 class gFunction(object):
@@ -339,6 +338,8 @@ class gFunction(object):
             Figure object (matplotlib).
 
         """
+        from ._mpl import plt
+
         # Configure figure and axes
         fig = _initialize_figure()
         ax = fig.add_subplot(111)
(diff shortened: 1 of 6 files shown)


MassimoCimmino:
Reading through the code, it seems to me the API has two objectives:

1. To follow-up on #308 and implement an API that uses the `Network` object to evaluate *g*-functions.
2. To facilitate the creation of `Network` (and `Pipe`) objects using static inputs and automate the evaluation of the required resistances.

For consistensy, my opinion is that (1.) should be addressed by implementing a `evaluate_g_function` method in the `Network` class, analogous to the method now implemented in the `Borefield` class.

My proposal for (2.) is to create a new class `GroundHeatExchanger` (or something similar) that encapsulates the cross-sectional geometry of the boreholes, evaluates thermal resistances, and includes methods to create `Pipe` or `Network` instances. Long-term, this could conveniently deal with #304 and also solve an issue of `Network` initialization where the multipole solution is evaluated `nBoreholes` times even when the geometry is the same. I have code from another project that could be adapted for implementation into pygfunction.

mitchute:
@MassimoCimmino thanks for your input. We can update this to adapt to your suggestions.

pygfunction/enums.py
```diff
@@ -0,0 +1,15 @@
+from enum import Enum, auto
+
+
+class GHEType(Enum):
+    BOREHOLE = auto()
+    BOREFIElD = auto()
+    NETWORK = auto()
```
mitchute, line 7: Needed this to break circular import issues.

pygfunction/gfunction.py
```diff
@@ -259,6 +257,13 @@ def __init__(self, boreholes_or_network, alpha, time=None,
         # Check the validity of inputs
         self._check_inputs()
 
+        # needed to break circular imports
+        from .solvers import (
+            Detailed,
+            Equivalent,
+            Similarities
+        )
```
mitchute, line 265: There's some lingering circular dependencies among the files. Needed to import these here to break the circular import problems.

pygfunction/gfunction.py
```diff
@@ -1236,11 +1241,11 @@ def _format_inputs(self, boreholes_or_network):
 
         """
         # Convert borehole to a list if a single borehole is provided
-        if isinstance(boreholes_or_network, Borehole):
+        if boreholes_or_network.ghe_type == GHEType.BOREHOLE:
```
mitchute, line 1244: Related to circular import issues

pygfunction/gfunction.py
```diff
@@ -1236,11 +1241,11 @@ def _format_inputs(self, boreholes_or_network):
...
             boreholes_or_network = [boreholes_or_network]
         # Check if a borefield or a network is provided as an input and
         # correctly assign the variables self.boreholes and self.network
-        if isinstance(boreholes_or_network, Network):
+        elif boreholes_or_network.ghe_type == GHEType.NETWORK:
```
mitchute, line 1248: It seemed to me like these should be chained if-else statements, not a series of `if` blocks, since they didn't `return` early.

pygfunction/gfunction.py
```diff
@@ -1531,238 +1538,3 @@ def uniform_temperature(boreholes, time, alpha, nSegments=8,
         boundary_condition=boundary_condition, options=options)
 
     return gFunc.gFunc
-
-
-def equal_inlet_temperature(
-        boreholes, UTubes, m_flow_borehole, cp_f, time, alpha,
-        kind='linear', nSegments=8, segment_ratios=segment_ratios,
-        use_similarities=True, disTol=0.01, tol=1.0e-6, dtype=np.double,
-        disp=False, **kwargs):
```
mitchute, line 1540: These are all tagged for deprecation. I removed them so I didn't have to continue to deal with circular import problems.

pygfunction/ground_heat_exchanger.py
```diff
@@ -0,0 +1,67 @@
...
+from .media import Fluid
+from .networks import Network
+from .pipes import get_pipes
+
+
+class GroundHeatExchanger:
+
+    def __init__(self,
+                 H: npt.ArrayLike,
+                 D: npt.ArrayLike,
```
mitchute, line 16: New GHE class. Constructor and `evaluate_g_function` arguments need a more thorough review.

pygfunction/networks.py
```diff
@@ -992,11 +999,11 @@ def _initialize_coefficients_connectivity(self):
...
         nMethods = 7    # Number of class methods
-        self._stored_coefficients = [() for i in range(nMethods)]
+        self._stored_coefficients = [() for _ in range(nMethods)]
         self._stored_m_flow_cp = [np.empty(self.nInlets)*np.nan
-                                  for i in range(nMethods)]
-        self._stored_nSegments = [np.nan for i in range(nMethods)]
-        self._stored_segment_ratios = [np.nan for i in range(nMethods)]
+                                  for _ in range(nMethods)]
+        self._stored_nSegments = [np.nan for _ in range(nMethods)]
+        self._stored_segment_ratios = [np.nan for _ in range(nMethods)]
```
mitchute, line 1006: unused args

pygfunction/pipes.py
```diff
@@ -3613,3 +3618,198 @@ def _Nusselt_number_turbulent_flow(Re, Pr, fDarcy):
     Nu = 0.125 * fDarcy * (Re - 1.0e3) * Pr / \
         (1.0 + 12.7 * np.sqrt(0.125*fDarcy) * (Pr**(2.0/3.0) - 1.0))
     return Nu
+
+
+def compute_R_fp(
+        pipe_type: PipeType, m_flow_borehole: float, r_in: Union[float, tuple, list],
+        r_out: Union[float, tuple, list], k_p: float, epsilon: float, fluid: Fluid) -> float:
+    if pipe_type in [PipeType.SINGLEUTUBE, PipeType.DOUBLEUTUBESERIES]:
```
mitchute, line 3626: Some worker functions

pygfunction/ground_heat_exchanger.py
```diff
@@ -0,0 +1,67 @@
...
+                 pipe_type: PipeType,
+                 pos: List[tuple],
+                 r_in: Union[float, tuple, list],
+                 r_out: Union[float, tuple, list],
+                 k_s: float,
+                 k_g: float,
+                 k_p: float,
+                 fluid_name: str,
+                 fluid_concentration_pct: float,
+                 m_flow_ghe,
```
mitchute, line 29: Would prefer to not have to pass this in to the constructor, but that's going to require some more work.

tests/gfunction_test.py
```diff
@@ -339,3 +339,126 @@ def test_gfunctions_UBWT_linearization(field, method, opts, expected, request):
...
+        #  'equivalent' solver - unequal segments - MIFT - coaxial annular inlet
+        ('single_borehole', 'MIFT', 'equivalent', 'unequal_segments', 'coaxial_annular_in', 0.05, np.array([6.10236427, 6.77069069, 6.95941276])),
+        ('single_borehole_short', 'MIFT', 'equivalent', 'unequal_segments', 'coaxial_annular_in', 0.05, np.array([4.06874781, 4.89701125, 5.19157017])),
+        ('ten_boreholes_rectangular', 'MIFT', 'equivalent', 'unequal_segments', 'coaxial_annular_in', 0.25, np.array([16.03433989, 21.18241954, 22.49479982])),
+        #  'equivalent' solver - unequal segments - MIFT - coaxial annular outlet
+        ('single_borehole', 'MIFT', 'equivalent', 'unequal_segments', 'coaxial_annular_out', 0.05, np.array([6.10236427, 6.77069069, 6.95941276])),
+        ('single_borehole_short', 'MIFT', 'equivalent', 'unequal_segments', 'coaxial_annular_out', 0.05, np.array([4.06874781, 4.89701125, 5.19157017])),
+        ('ten_boreholes_rectangular', 'MIFT', 'equivalent', 'unequal_segments', 'coaxial_annular_out', 0.25, np.array([16.03433989, 21.18241954, 22.49510883])),
+    ])
```
mitchute, line 380: Some tests to verify things are nominally functional. Expected values pulled from existing tests from above.

mitchute (commented):
@MassimoCimmino thanks for your feedback. These updates are a first pass at addressing your suggestions.

PR #325, opened 2025-06-29
Create add function for Borefield Class

wouterpeere:
Hi @MassimoCimmino ,

I thought it would be useful to have a function to add different borefields together.

Best,
Wouter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 2.4 (in development)
 
+### New features
+
+* [Pull Request 325](https://github.com/MassimoCimmino/pygfunction/pull/325) - Borefields and boreholes can now be concatenated using the `+` operator, e.g. using `new_field = field_1 + field_2`.
+
 ### Other changes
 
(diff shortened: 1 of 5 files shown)


MassimoCimmino:
Thanks. This was also on my to-do list.

We might want to implement also `__radd__` for the case `borehole + borefield`. One thing that delayed me on this feature is how to do the operation `borehole1 + borehole2`. Perhaps you have an idea there.

Following this, we could explore adding methods to the borefield class such as `.translate` and `.rotate` to have composable borefields.

MassimoCimmino:
Thanks @wouterpeere.

Your implementation of `Borehole.__add__(Borefield)` always put the borehole at the end of the borefield. Calling `borehole + field` would have the expected result of putting the borehole at the start of the borefield.

I fixed it by implementing both `Borehole.__radd__` and `Borefield.__radd__`. This as the added bonus of also handling the cases `list + borehole` and `list + field`.

Could you doublecheck the implementation before I merge?

wouterpeere:
@MassimoCimmino , everything looks good to me!

(I tried if the __rand__ was really needed, since I'm not that familiar with it, but everything is fine!)

